PR #16, opened 2018-08-22
add collections chapter

japaric:
This is what I proposed to include in this section when we discussed the outline
some time ago.

I have used `alloc` and `heapless` as examples but what I mainly wanted to cover
is the heap allocated, relocatable collections approach vs the fixed capacity
collections approach and their different trade-offs.

@rust-embedded/resources Thoughts on this topic?

(Also I replaced "dynamic data structures" with "collections" because it was
shorter and I believe most people will recognize the name / concept)

### DIFF
--- a/src/collections/collections.md
+++ b/src/collections/collections.md
@@ -1,0 +1,258 @@
+# Collections
+
+Eventually you'll want to use dynamic data structures (AKA collections) in your
+program. `std` provides a set of common collections: [`Vec`], [`String`],
+[`HashMap`], etc. All the collections implemented in `std` use a global dynamic
+memory allocator (AKA the heap).
+
+[`Vec`]: https://doc.rust-lang.org/std/vec/struct.Vec.html
+[`String`]: https://doc.rust-lang.org/std/string/struct.String.html
+[`HashMap`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
+
+As `core` is, by definition, free of memory allocations these implementations
+are not available there, but they can be found in the *unstable* `alloc` crate
+that's shipped with the compiler.
+
+If you need collections, a heap allocated implementation is not your only
+option. You can also use *fixed capacity* collections; one such implementation
+can be found in the [`heapless`] crate.
+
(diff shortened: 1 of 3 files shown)


jamesmunns:
Looked like a network error, restarted travis

src/collections/collections.md
```diff
@@ -0,0 +1,258 @@
...
+allocator for anything else then an OOM condition is impossible. Instead, you'll
+have to deal with collections running out of capacity on a case by case basis.
+That is you'll have deal with *all* the `Result`s returned by methods like
+`Vec.push`.
+
+OOM failures can be harder to debug than say `unwrap`-ing on all `Result`s
+returned by `heapless::Vec.push` because the observed location of failure may
+*not* match with the location of the cause of the problem. For example, even
+`vec.reserve(1)` can trigger an OOM if the allocator is nearly exhausted because
+some other collection was leaking memory (memory leaks are possible in safe
```
Nicoretti: Out of curiosity + lack of knowledge: How would safe rust code, which does produce a memory leak, look like? 
Or was this note under the premiss, that safe rust code interacts with (e.g. collections) which might contain unsafe code with memory leaks?

src/collections/collections.md
```diff
@@ -0,0 +1,258 @@
+# Collections
```
Nicoretti: From somebody who isn't as experienced as you are:
Great read! For me all key information is there.

src/collections/collections.md
japaric: > How would safe rust code, which does produce a memory leak, look like?

`mem::forget(vec![1])` leaks memory and it's safe. `mem::forget` used to be `unsafe` until we noticed it was also possible to leak memory by creating reference cycles using the safe API of the `Rc` abstraction.

The day we made `forget` safe we also declared that depending on destructors running for memory safety is an anti-pattern (see the old fork-join API that was in the standard library; it never made into stable and was removed). This chain of events is known as the "leakpocalyse".

japaric:
ping @rust-embedded/resources this is a friendly reminder to review this.

PR #17, opened 2018-08-24
[Draft]  no_std section -- Feedback needed

Nicoretti:
Hi @all,

I started drafting something for the no_std section, but I surely lack the clear picture who will
be the main/target audience for the book. So If you can give me some feedback (keywords, hints) whats missing and whats too much
e.g. if we assume people to know stuff like the **"Static" vs. Dynamic memory** part etc.
I'll be happy to work it in.

### DIFF
--- a/src/intro/no-std.md
+++ b/src/intro/no-std.md
@@ -1,5 +1,74 @@
 # A `no_std` Rust Environment
 
-> ❌: **This section should cover:**
->
-> * what #[no_std] means: no I/O (filesystems), allocation, etc.
+As mentioned earlier, this book is all about "bare metal" programming. Standard libraries in turn usually
+require some sort of system (os) integration and higher level functionality, on top of which they implement their functionality.
+Such functionalities are for example I/O (files, networking), Threads, and Heap (dynamic memory allocation).
+
+But don't worry, this does not mean you have to code everything from scratch.
+In an `no_std` environment you still will be able to use the [libcore](https://doc.rust-lang.org/core/),
+if you provide a few basic functions (Either by pulling in in the [rlibc](https://crates.io/crates/rlibc) crate,
+or by implementing them by yourself).
+
+See also:
+* [FAQ](https://www.rust-lang.org/en-US/faq.html#does-rust-work-without-the-standard-library)
+* [RFC-1184](https://github.com/rust-lang/rfcs/blob/master/text/1184-stabilize-no_std.md)
+
+
+## Does this mean I can't use dynamic memory or any kind of I/O?
+No it doesen't! It just means you need to do some work in order to bootstrap it.
+For the cortex-m based setup this book is based on there are even some creates
+which will simplify some of bootstraping.
+
+So if you curios check out this crates:
+* [cortex-m-quickstart](https://crates.io/crates/cortex-m-quickstart)
+* [cortex-m-rt](https://crates.io/crates/cortex-m-rt),
+* [cortex-m-log](https://crates.io/crates/cortex-m-log)
+* [cortex-m-semihosting](https://crates.io/crates/cortex-m-semihosting)
+* [alloc-cortex-m](https://crates.io/crates/alloc-cortex-m)
+* [linked_list_allocator](https://crates.io/crates/linked_list_allocator).
+
+But before you start adding those crates and start allocating consider reading the next section
+`"Static" vs Dynamic memory alloctation` beforehand.
+
+## "Static" vs. Dynmaic memory allocation
+First let us define what we mean with when using the terms "Static" and Dynamic memory allocation in this section.
+
+"Static":
+```
+Means all memory that can be reserved and calculated at compile time,
+assuming no computiation based on dynamic inputs is done (e.g. recursion based on "user input").
+```
+Pros:
+* Memory usage can be known upfront
+* Faster than dynamic memory allocation
+* Time used for "allocation" predictable
+
+Cons:
+* Maxias need to be known before hand
+* Code will be less dynamic even if enough memory would be available
+
+Dynamic:
+```
+Means all heap allocations in this context, assuming a heap was set up.
+```
+
+Pros:
+* Code can be more dynamic
+* Freed memory can be reused
+
+Cons:
+* Overall memory usage hard to predict
+* Memory fragmentation/waste
+* Runtime impact, depending on the implementation and strategies of the allocator memory allocations
+  can take a significant amount of time.
+
+In embeded especially in embedded rt systems its often benefical if not required to have a predictable runtime
+behaviour. Meaning the more of the system behaviour you can verify and guarentee before its even running the better.
+Dynamic memory allocation can complicate this quite a bit.
+This is why such systems do not use dynamic memory alloction, or only use an known amount of allocations right
+at the startup/init.
+
+
+> ❌: For more details on best practises / dev patterns in embedded environment see XYZ
+
+>     Should we add chapter for that (e.g. no recursion etc.)
--- a/src/intro/no-std.md
+++ b/src/intro/no-std.md
@@ -1,74 +1,34 @@
 # A `no_std` Rust Environment
 
-As mentioned earlier, this book is all about "bare metal" programming. Standard libraries in turn usually
-require some sort of system (os) integration and higher level functionality, on top of which they implement their functionality.
-Such functionalities are for example I/O (files, networking), Threads, and Heap (dynamic memory allocation).
+The term Embedded Programming is used for a wide range of different classes of programming.
+Ranging from programming 8 Bit MCUs (like [ST72325xx](https://www.st.com/resource/en/datasheet/st72325j6.pdf)) with just a few KB of RAM and ROM, up to systems like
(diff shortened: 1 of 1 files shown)


src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
 # A `no_std` Rust Environment
 
-> ❌: **This section should cover:**
->
-> * what #[no_std] means: no I/O (filesystems), allocation, etc.
\ No newline at end of file
+As mentioned earlier, this book is all about "bare metal" programming. Standard libraries in turn usually
```
therealprof: Is it? I wouldn't say so, embedded rust is meant to work a lot more systems including those with embedded OSes or even Linux running on them (like the Raspberry Pi). I think the essence is that the latter is not actually required to use the resources provided by embedded systems.

src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
...
+require some sort of system (os) integration and higher level functionality, on top of which they implement their functionality.
+Such functionalities are for example I/O (files, networking), Threads, and Heap (dynamic memory allocation).
```
therealprof: Not quite accurate. Embedded systems are really all about I/O, just a different kind and level of I/O than typical developers are thinking of.

src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
...
+
+But don't worry, this does not mean you have to code everything from scratch.
+In an `no_std` environment you still will be able to use the [libcore](https://doc.rust-lang.org/core/),
+if you provide a few basic functions (Either by pulling in in the [rlibc](https://crates.io/crates/rlibc) crate,
```
therealprof: I didn't even know rlibc existed. ;) Looks pretty much dead to me. Very little is needed to compile a program for a MCU nowadays, mostly linker magic.

src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
...
+* [cortex-m-log](https://crates.io/crates/cortex-m-log)
+* [cortex-m-semihosting](https://crates.io/crates/cortex-m-semihosting)
+* [alloc-cortex-m](https://crates.io/crates/alloc-cortex-m)
+* [linked_list_allocator](https://crates.io/crates/linked_list_allocator).
+
+But before you start adding those crates and start allocating consider reading the next section
+`"Static" vs Dynamic memory alloctation` beforehand.
+
+## "Static" vs. Dynmaic memory allocation
+First let us define what we mean with when using the terms "Static" and Dynamic memory allocation in this section.
```
therealprof: ~with~

src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
...
+So if you curios check out this crates:
+* [cortex-m-quickstart](https://crates.io/crates/cortex-m-quickstart)
+* [cortex-m-rt](https://crates.io/crates/cortex-m-rt),
+* [cortex-m-log](https://crates.io/crates/cortex-m-log)
+* [cortex-m-semihosting](https://crates.io/crates/cortex-m-semihosting)
+* [alloc-cortex-m](https://crates.io/crates/alloc-cortex-m)
+* [linked_list_allocator](https://crates.io/crates/linked_list_allocator).
+
+But before you start adding those crates and start allocating consider reading the next section
+`"Static" vs Dynamic memory alloctation` beforehand.
```
therealprof: Why Static in quotes?

src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
...
+
+## "Static" vs. Dynmaic memory allocation
+First let us define what we mean with when using the terms "Static" and Dynamic memory allocation in this section.
+
+"Static":
+```
+Means all memory that can be reserved and calculated at compile time,
```
therealprof: You probably mean that the memory locations can be calculated and assigned at compile time?

src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
...
+assuming no computiation based on dynamic inputs is done (e.g. recursion based on "user input").
```
therealprof: computation

therealprof: ... and well, it is it is definitely possible to computation based on dynamic inputs (every input is dynamic, otherwise you'd just be running a completely pre-determined schedule), just the size of the data structures cannot change at runtime.

src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
...
+```
+Pros:
+* Memory usage can be known upfront
+* Faster than dynamic memory allocation
```
therealprof: Typically yes but not necessarily always.

src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
...
+## "Static" vs. Dynmaic memory allocation
+First let us define what we mean with when using the terms "Static" and Dynamic memory allocation in this section.
+
+"Static":
+```
+Means all memory that can be reserved and calculated at compile time,
+assuming no computiation based on dynamic inputs is done (e.g. recursion based on "user input").
+```
+Pros:
+* Memory usage can be known upfront
```
therealprof: Well, that's the prerequisite. The pro would probably be that it is possible to tell whether the system actually has enough memory to fit all the static data.

src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
...
+Pros:
+* Code can be more dynamic
+* Freed memory can be reused
+
+Cons:
+* Overall memory usage hard to predict
+* Memory fragmentation/waste
+* Runtime impact, depending on the implementation and strategies of the allocator memory allocations
+  can take a significant amount of time.
+
```
therealprof: The biggest con of all is: If you cannot fulfil a memory allocation you either have to deal with it in a clever way or the system will panic.

src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
...
+In embeded especially in embedded rt systems its often benefical if not required to have a predictable runtime
+behaviour. Meaning the more of the system behaviour you can verify and guarentee before its even running the better.
```
therealprof: guarantee

src/intro/no-std.md
```diff
@@ -1,5 +1,74 @@
...
+Dynamic memory allocation can complicate this quite a bit.
+This is why such systems do not use dynamic memory alloction, or only use an known amount of allocations right
+at the startup/init.
```
therealprof: Well, actually they very often do when written in other languages. I would consider it a benefit of Rust that it is very often not necessary to use dynamic allocations in Rust and if you do not (and stay away from unsafe code) Rust will actually guarantee that your program will run properly; static allocations in other languages do not provide that particular property, e.g. even if you use statically allocated data structures in other languages there's no guarantee that the code using them will stay within the specified boundaries of the data structure.

src/intro/no-std.md
Nicoretti: Because one could say even if you do not have and/or use the heap memory/stack space will be "allocated" during runtime.

src/intro/no-std.md
Nicoretti: Partially true, one needs to know the worst case (the worst case which one wants to be able to handle). It might be e.g. totally legit to reserve a static amount of buffer for ethernet and just document that if traffic exceeds XYZ packets will be dropped.

src/intro/no-std.md
Nicoretti: agree!

src/intro/no-std.md
Nicoretti: Very true, expirienced it first hand xD. 
I think this is a good point we could make in the "Who Embedded Rust is for", isn't it?

src/intro/no-std.md
Nicoretti: Not only, for example if you do not use things like recursion you can determine call graphs and calculate the call stack sizes. The the [Diab Compiler](https://www.embedded-tools.de/sites/default/files/pdf/wr_diab_compiler_options_reference_5.9.3-R2.pdf) for example does something like this (he does a link time stack usage analysis).
(see option: -Xstack-usage)

src/intro/no-std.md
Nicoretti: yes you are right, I will rephrase or remove that.

src/intro/no-std.md
thejpster: Yeah I'd change that to `are for example file I/O, network sockets, threads and dynamic memory allocation`.

Nicoretti:
@japaric thx for the reference to your write up. Your post definitely contains additional information which is very valuable.I agree that we should not duplicate information, so I would be happy to merge the two texts together. I'm just not sure about the timeline (do we have a specific **date to land this?**). 

But can propose this:
I'll start on working on merging the texts, if you need to land it, before I can provide the update you can use the current state and I'll create a PR once I've finished the merge.

Info:
If you need the merge quickly you should consider delegating the work to somebody else because I'm not sure I can start working on the follow up before 09.09.2018.

jamesmunns:
Hey @Nicoretti, I think we are okay to merge this as-is today, and if you have a chance later to merge in topics from Japaric's work, that can be a follow-on PR.

Let me know if you're okay with this, and I can trigger the merge through bors.

PR #22, opened 2018-09-11
Add additional details to the no_std section

Nicoretti:
* information about the stdlib runtime
* details about the no_std attribute itself
* a more detailed summary
* a table for comparision of supported and unsupported features of std vs. no_std environments

### DIFF
--- a/src/intro/no-std.md
+++ b/src/intro/no-std.md
@@ -19,15 +19,33 @@ special HW or I/Os. Overall it feels like coding in a special-purpose PC environ
 In a bare metal environment there will be no high level OS running and hosting our code.
 This means there will be no primitives, which means there's also no standard library by default.
 By marking our code with `no_std` we indicate that our code is capable of running in such an environment.
-This means the rust [stdlib](https://doc.rust-lang.org/std/) and dynamic memory allocation can't be used by such code.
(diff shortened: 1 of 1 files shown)


Nicoretti:
Hi @all,

I have tried to work in missing details into the `no_std` section mentioned by @japaric  (from [embedonomicon](https://rust-embedded.github.io/embedonomicon/smallest-no-std.html#what-does-no_std-mean)) would be nice if you could have a look at it.

src/intro/no-std.md
```diff
@@ -23,12 +23,31 @@ This means the rust [stdlib](https://doc.rust-lang.org/std/) and dynamic memory
 However, such code can use [libcore](https://doc.rust-lang.org/core/), which can easily be made available
 in any kind of environment by providing just a few symbols (for details see [libcore](https://doc.rust-lang.org/core/)).
 
+### The stdlib Runtime
+As mentioned before using stdlib requires some sort of system integration, but this is not only because
```
jamesmunns: Small nitpick, you might want to use `libstd` across this page rather than `stdlib`. I think the former is more common in Rust nomenclature

Nicoretti: good point, thx -> I'll update the PR

therealprof:
@Nicoretti Amending or rebasing and force pushing is fine, GitHub handles that very well.

PR #23, opened 2018-09-12
[WIP] Adding C interop sections

spacekookie:
*This is still WIP, missing a bunch of stuff.*

This PR adds sections to the book about how to embed Rust code into C/C++ projects and how to interact between Rust and C, transforming data into the appropriate formats.

### DIFF
--- a/src/interoperability/interoperability.md
+++ b/src/interoperability/interoperability.md
@@ -1,3 +1,27 @@
 # Interoperability
 
-> ❌: This section has not yet been written. Please refer to [embedded-wg#48](https://github.com/rust-lang-nursery/embedded-wg/issues/48) for discussion of this section.
+Iteroperability between Rust and C code is always dependant 
+on transforming data between the two languages.
+For this purposes there are two dedicated modules 
+in the `stdlib` called `std::ffi` and `std::raw`.
+
+For `#![no_std]` environments, there is also `core::raw` that
+covers all the primitive types.
+
+------------------------------------------
+| Rust type | Intermediate | C type       |
+|-----------|--------------|--------------|
+| String    | CString      | *char        |
+| &str      | CStr         | *const char  |
+| ()        | c_void       | void         |
+| u32 | u64 | c_uint       | unsigned int |
+| etc       | ...          | ...          |
+
+Primitive types can simply be cast to Rust types.
+
+```rust
+unsafe fn foo(num: u32) {
+    let c_num: c_uint = num;
+    let r_num: u32 = c_num;
+}
+```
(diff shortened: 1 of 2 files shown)


src/interoperability/interoperability.md
```diff
@@ -1,3 +1,27 @@
 # Interoperability
 
-> ❌: This section has not yet been written. Please refer to [embedded-wg#48](https://github.com/rust-lang-nursery/embedded-wg/issues/48) for discussion of this section.
\ No newline at end of file
+Iteroperability between Rust and C code is always dependant 
+on transforming data between the two languages.
+For this purposes there are two dedicated modules 
+in the `stdlib` called `std::ffi` and `std::raw`.
```
andre-richter: Do you plan to elaborate more on what ffi and raw do?

andre-richter:
Thanks for starting this. I left a little inline comment.

src/interoperability/rust-with-c.md
```diff
@@ -1 +1,105 @@
...
+### `#[no_mangle]`
+
+The Rust compiler mangles symbol names differently than native code linkers expect.
+As such, any function that Rust exports to be used outside of Rust needs to be told
+not to be mangled by the compiler.
+
+### `extern "C"`
+
+By default, any function you write in Rust will use the 
+Rust ABI (which is also no stabilised).
```
posborne: sp: also _not_ stabilized.

src/interoperability/interoperability.md
posborne: Making these hyperlinks to the docs would probably be handy.

src/interoperability/interoperability.md
```diff
@@ -1,3 +1,27 @@
...
+------------------------------------------
+| Rust type | Intermediate | C type       |
+|-----------|--------------|--------------|
+| String    | CString      | *char        |
+| &str      | CStr         | *const char  |
+| ()        | c_void       | void         |
+| u32 | u64 | c_uint       | unsigned int |
+| etc       | ...          | ...          |
+
+Primitive types can simply be cast to Rust types.
```
posborne: Maybe instead of "cast" the wording "converted into" would be more appropriate for the example shown.  In this case, `c_uint` and `u32` end up being the same type but that won't always be the case.

Just thinking the audience might react negatively to the idea of a "cast" being required (which often implies unsafety).

spacekookie:
Apologies for the long delay before getting back to this! 

I fixed the marked sections in the existing text, but I'll try to finish the rest of it by the end of the week!

PR #25, opened 2018-09-18
initial FAQ

japaric:
I'm not entirely sure if the FAQ should live in the book or as a separate
thing (e.g. docs.rust-embedded/faq) but I think it's best to write down
something now and move it later, if desired

cc rust-embedded/wg#13 (maybe we should move that issue here?)

r? @rust-embedded/resources (anyone)

### DIFF
--- a/src/unsorted/faq.md
+++ b/src/unsorted/faq.md
@@ -1,0 +1,167 @@
+# Frequently Asked Questions
+
+- [Does Rust support my device][#does-rust-support-my-device]
+
+## Does Rust support my device?
+
+### Short answer
+
+As of 2018-09-18 the Rust compiler supports cross compiling to these embedded
+architectures (see `rustup target list`):
+
+- ARM Cortex-M (since 1.27)
+  - `thumbv6m-none-eabi`, Cortex-M0
+  - `thumbv7m-none-eabi`, Cortex-M3
+  - `thumbv7em-none-eabi`, Cortex-M4 and Cortex-M7
+  - `thumbv7em-none-eabihf`, Cortex-M4F and Cortex-M7F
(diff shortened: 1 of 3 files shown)


src/unsorted/faq.md
```diff
@@ -0,0 +1,167 @@
...
+  - `thumbv7m-none-eabi`, Cortex-M3
+  - `thumbv7em-none-eabi`, Cortex-M4 and Cortex-M7
+  - `thumbv7em-none-eabihf`, Cortex-M4F and Cortex-M7F
+- ARM Cortex-R (1.30-beta)
+  - `armebv7r-none-eabi`, big endian Cortex-R4 and Cortex-R5
+  - `armebv7r-none-eabihf`, big endian Cortex-R4F and Cortex-R5F
+  - `armv7r-none-eabi`, little endian Cortex-R4 and Cortex-R5
+  - `armv7r-none-eabihf`, little endian Cortex-R4F and Cortex-R5F
+- ARM Linux
+  - ARMv5TE, ARMv6, ARMv7 and ARMv8 in GNU and MUSL flavors and in soft float
```
thejpster: Can we give some examples here, like we did above with the Cortex-M<foo>? Like the ARM11 is an ARMv6 (relevant, as it's on a Raspberry Pi Zero/1), while an ARM926EJ-S is ARMv5, Cortex-A8 is ARMv7-A (sic) and Cortex-A53 is ARMv8-A (sic).

thejpster: Also worth noting that that ARMv4 (ARM7) is not supported. I mean, the most recent model came out 17 years ago, but some may be interested in e.g. Dreamcast homebrew.

japaric: Done

PR #26, opened 2018-09-18
Optimizations: the speed size tradeoff

japaric:
cc #11

another unsorted topic

r? @rust-embedded/resources


japaric:
Note to self: cover the fact that, today, fat LTO is better than thin LTO.

therealprof:
NB: Looking at the overall binary size is certainly a good idea. But for better analysis of the effect of optimisation and individual function sizes (or inlining) we might want to mention `cargo-bloat`.

PR #27, opened 2018-09-24
new section: panicking

japaric:
this probably should go in the second chapter but I'm putting it in the unsorted
section until #20 lands

Is there anything else that should be mentioned in this section?

r? @rust-embedded/resources


japaric:
I forgot to mention but I have moved the section under the second chapter now that #20 has landed. I have also added an example.

PR #28, opened 2018-09-24
new section: exceptions

japaric:
I was pleasantly surprised to find out that QEMU properly emulates the system
timer peripheral.

r? @rust-embedded/resources

### DIFF
--- a/src/start/exceptions.md
+++ b/src/start/exceptions.md
@@ -1,7 +1,256 @@
 # Exceptions
 
-> **TODO** Cover statically registering exception handlers using the `exception`
-> attribute; the default exception handler (`DefaultHandler`); and safe local
-> `static mut` variables (non-reentrant functions). Also cover (perhaps in a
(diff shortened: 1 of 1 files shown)


japaric:
Since the examples use semihosting we should briefly cover semihosting in a previous section.

src/start/exceptions.md
```diff
@@ -1,7 +1,256 @@
...
+called by software. Following the previous example, the statement `SysTick();`
+would result in a compilation error.
+
+This behavior is pretty much intended and it's required to provide a feature:
+`static mut` variables declared *inside* `exception` handlers are *safe* to use.
+
+``` rust
+#[exception]
+fn SysTick() {
+    static mut COUNT: u32 = 0;
```
therealprof: Have you ever tried slotting something between a `static mut` declaration and the`fn` like a `use` statement? That a real fun exercise and head scratcher. ;)

PR #36, opened 2018-10-02
WIP: Portability Chapter

jamesmunns:
Not much here yet, help wanted!

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -32,6 +32,7 @@ more information and coordination
 - [Static Guarantees](./static-guarantees/static-guarantees.md)
     <!-- TODO: Define Sections -->
 - [Portability](./portability/portability.md)
+    - [The Trait System](./portability/traits.md)
     <!-- TODO: Define more sections -->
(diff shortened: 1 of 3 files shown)


jamesmunns:
@therealprof I am going to close this, but I would say this PR has the following redeeming features I would like to see reintroduced to the `embedded-hal` chapter:

* Examples (we've already discussed)
* Commentary on why these traits are useful, (N*M vs N+M implementation, where N is # of chip/board support crates, and M is the # of driver traits).

Additionally, we may want something like [this image](https://screenshots.firefox.com/RviflD6enGaYiZRm/ferrous-systems.github.io), to demonstrate how this all fits together, and we may want to talk about the different components in the ecosystem somewhere:

* Peripheral Access Crates
* HAL Implementations
* Board Support Crates
* `embedded-hal`
* Driver Crates
* `no_std` libraries/crates (such as `heapless`, `menu`, etc.)

jamesmunns:
Moving my comment to an issue, easier to track.

therealprof:
Sounds good.

PR #37, opened 2018-10-02
fix arch package name for qemu-system-arm

rmsthebest:
see: https://wiki.archlinux.org/index.php/QEMU#Installation

### DIFF
--- a/src/intro/install/linux.md
+++ b/src/intro/install/linux.md
@@ -70,7 +70,7 @@ $ sudo dnf install \
 ``` console
 $ sudo pacman -S \
   arm-none-eabi-gdb \
-  qemu-system-arm
+  qemu-arch-extra
 
 $ # install openocd from the AUR -- https://aur.archlinux.org/packages/openocd/
 ```


andre-richter:
Not an Arch user here. From what I can see, `qemu-arch-extra` is an add-on package to the `qemu` package. The latter is not explicitly mentioned in the command. Will it be pulled in as a dependency automatically?

rmsthebest:
> Not an Arch user here. From what I can see, `qemu-arch-extra` is an add-on package to the `qemu` package. The latter is not explicitly mentioned in the command. Will it be pulled in as a dependency automatically?

qemu is a dependency

sudo pacman -S qemu-arch-extra
[sudo] password for cameron:
resolving dependencies...
looking for conflicting packages...

Packages (11) jemalloc-1:5.1.0-1  libcacard-2.6.1-1  numactl-2.0.12-1  **qemu-3.0.0-2**
              seabios-1.11.0-1  snappy-1.1.7-1  usbredir-0.7.1-1  vde2-2.3.2-11
              virglrenderer-0.7.0-1  vte3-0.54.1-1  qemu-arch-extra-3.0.0-2

Total Download Size:    90.25 MiB
Total Installed Size:  542.41 MiB

PR #38, opened 2018-10-03
Insert /book/ into link URL

therealprof:
Hopefully this will fix the broken URL for good. However I'm wondering whether we shouldn't make it absolute instead so it would also render correctly in the preview on GitHub...

r? @rust-embedded/resources


therealprof:
Did the trick. Thanks.

PR #45, opened 2018-10-14
First pass at "C with Rust"

jamesmunns:
Took a first pass at this, I feel like there are probably a few places where I could use more/better examples or links to external information.

Please feel free to review and suggest changes.

Companion to @spacekookie's work in #23.


jamesmunns:
bors r=ryankurte

PR #49, opened 2018-10-16
First draft at concurrency chapter

### LINKED ISSUE #7: Book section: Concurrency (draft)
_From @japaric on July 17, 2018 14:34_

This section should cover concurrency in embedded systems / microcontrollers. How the Send / Sync traits can be used with interrupts and static variables to achieve memory safe concurrency in those systems. We start with the simple case of interrupt handlers, static variables and critical sections and then show that a framework / DSL like Real Time for The Masses can built using those basic blocks.

If Tock userland apps (not the kernel) are working on stable then we can include a subsection about that as well as it demoes the multiprocess architecture. Same thing with freertos-rs which demoes a more traditional threading system (I think).

_Copied from original issue: rust-embedded/wg#124_

adamgreig:
I've mostly focused on building up to safely sharing peripherals, without very much focus on RTFM or higher level things like an RTOS, since I don't think there's very much we can usefully say about them at the moment. Very happy to be told I'm wrong on that front though!

Closes #7.

r? @rust-embedded/resources

### DIFF
--- a/src/concurrency/concurrency.md
+++ b/src/concurrency/concurrency.md
@@ -1,3 +1,537 @@
 # Concurrency
 
-> ❌: This section has not yet been written. Please refer to [rust-embedded/book#7](https://github.com/rust-embedded/book/issues/7) for discussion of this section.
+Concurrency happens whenever different parts of your program might execute
+at different times or out of order. In an embedded context, this includes:
+
+* interrupt handlers, which run whenever the associated interrupt happens,
+* various forms of multithreading, where your microprocessor regularly swaps
+  between parts of your program,
+* and in some systems, multiple-core microprocessors, where each core can be
+  independently running a different part of your program at the same time.
+
+Since many embedded programs need to deal with interrupts, concurrency will
+usually come up sooner or later, and it's also where many subtle and difficult
+bugs can occur. Luckily, Rust provides a number of abstractions and safety
+guarantees to help us write correct code.
+
+## No Concurrency
+
+The simplest concurrency for an embedded program is no concurrency: your
+software consists of a single main loop which just keeps running, and there
+are no interrupts at all. Sometimes this is perfectly suited to the problem
+at hand! Typically your loop will read some inputs, perform some processing,
+and write some outputs.
+
+```rust
+#[entry]
+fn main() {
+    let peripherals = setup_peripherals();
+    loop {
+        let inputs = read_inputs(&peripherals);
+        let outputs = process(inputs);
+        write_outputs(&peripherals, outputs);
+    }
+}
+```
+
+Since there's no concurrency, there's no need to worry about sharing data
+between parts of your program or synchronising access to peripherals. If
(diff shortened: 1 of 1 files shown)


andre-richter:
This is high quality content. Hats off!

I am wondering about one point: The mutex-inside-critical-section example confuses me a bit. In a single-core scenario, it doesn't make much sense, because a critical section is already enough to keep the only contender for the shared data away? In a multi-core example, we could still get weird behavior if one core currently holds the lock when the other core in an ISR tries to lock it. Maybe I am missing something here?

adamgreig:
Thanks!

In the single-core case, the mutex is really only ensuring you are in fact inside a critical section - it doesn't have its own state or concept of being locked or any other overhead. Really all it's doing is only letting you borrow its contents if you can give it a `CriticalSection` token. We need it rather than just using the shared data directly only because the shared data would not be Sync (e.g. a `Cell` or `RefCell`) but the `Mutex<Cell>` _is_ Sync, so can be placed in statics.

In a multi-core scenario the `cortex_m::interrupt::Mutex` is unsound as far as I can see...

andre-richter:
I was just asking because the introduction speaks about multi-core. Would it help to set the scope for each of the examples (single or multi or both) so that the reader can differentiate?

adamgreig:
Good idea, I'll add a note that the `cortex_m::interrupt::Mutex` (and in general relying on critical sections for exclusive access) only applies on single core systems.

adamgreig:
@andre-richter I've added some notes about multiple cores. Good observation: it's lead to needing to also note that `cortex_m::interrupt::Mutex` is not actually safe on multi-core systems!

src/concurrency/concurrency.md
```diff
@@ -1,3 +1,537 @@
...
+
+Whew! This is safe, but it is also a little unwieldy. Is there anything else
+we can do?
+
+## RTFM
+
+One alternative is the [RTFM framework], short for Real Time For the Masses. It
+uses interrupt priorities to statically ensure that shared resources are always
+accessed safely, without requiring the overhead of always entering critical
+sections and using reference counting (as in `RefCell`). This has a number of
```
therealprof: No reference "counting" here.

adamgreig: I think it's fair to call what `RefCell` does "reference counting": it literally counts how many `&` and `&mut` have been taken to dynamically enforce the borrow rules. It's not the same as a reference counted pointer (like `Rc`), which counts references to know when to drop a value from the heap, but it's still reference counting.

therealprof: After reading the code I agree with your assessment about the implementation details which where not clear to me before. However I still think it is confusing to mention "reference counting" as an implementation detail here and would prefer to use a different phrase such as "maintaining the borrow state" to refer to the overhead incurred by using `RefCell`.

andre-richter:
Shall we bors this @adamgreig ?

PR #54, opened 2018-10-19
Fix broken link

BenBergman:


thejpster:
If we're fixing this one, could we canonicalise all of the URLs so they're in the same format?

Looking through I can see:

```
src/start/hardware.md
41:[previous section on QEMU]: /start/qemu.md
105:[verify]: /intro/install/verify.md

README.md
28:[MIT License]: ./LICENSE-MIT

src/intro/install/verify.md
48:[next section]: /install/hardware.html
65:[udev rules]: /install/linux.html
```

Ideally the links would work both on Github.com, and in the rendered book.

therealprof:
@thejpster Amen to that. Do you have any idea what the links need to look like to work in both cases?

thejpster:
Take a look at https://github.com/thejpster/book/blob/testing_links/src/intro/install/verify.md.
It contains:

```
The contents may not match exactly but you should get the last line about
breakpoints and watchpoints. If you got it terminate the OpenOCD process and
move to the a [test1] b [test2] c [test3] d [test4] e [test5] f [test6].

[test1]: /intro/hardware.html
[test2]: ./intro/hardware.html
[test3]: /intro/hardware.md
[test4]: ./intro/hardware.md
[test5]: intro/hardware.md
[test6]: ../hardware.md
```

Only test1, test3 and test6 work in `mdbook serve`. Only test6 works on Github. Therefore I propose all URLs are made relative, with a `.md` extension, and no leading `./` where the file is in the same directory.

japaric:
Thanks for the original PR, @BenBergman. All the links have been updated to be relative in #68, which also fixed this broken link.

PR #58, opened 2018-10-20
Attempt to fix mdbook v0.2.2 issue by using v0.2.1

therealprof:
v0.2.2 was just released but does not include the relevant packages yet so install.sh fails


therealprof:
bors try

PR #71, opened 2018-11-04
concurrency: small tweaks to the atomics section

japaric:
r? @adamgreig

### DIFF
--- a/src/concurrency/concurrency.md
+++ b/src/concurrency/concurrency.md
@@ -192,13 +192,12 @@ main thread without disabling interrupts. When possible, this is a better
 solution — but it may not be supported on your platform.
 
 A note on [`Ordering`]: this affects how the compiler and hardware may reorder
-instructions, and also has consequences on cache visibility. For simple atomic
-operations like incrementing a counter, where we are not synchronising any
-other tasks on the counter, `Relaxed` is sufficient, and will have the best
-performance on typical embedded platforms. Stricter ordering will cause the
-compiler to emit memory barriers around the atomic operations; depending on
-what you're using atomics for you may or may not need this! The precise
-details of the atomic model are complicated and best described elsewhere.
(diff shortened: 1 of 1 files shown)


adamgreig:
Oops, I should not have put them as `static mut` in the first place! Thanks.

I'm not _completely_ sure `Relaxed` is always sufficient on single-core platforms (what about Cortex-M7 with its data cache that needs `DMB` etc when DMA might read it? what about compiler re-ordering instructions around a lock atomic in the face of interrupts triggering?). I get that `Relaxed` is always fine for a counter (as in this example), but I thought that would be true even on multi-core systems.

Anyway I think the new text is fine too and really anyone needing to know about this probably needs to read more than this book, so.. lgtm.

src/concurrency/concurrency.md
```diff
@@ -174,30 +174,30 @@ fn main() -> ! {
         if state && !last_state {
             last_state = state;
             // Use `fetch_add` to atomically add 1 to COUNTER
-            unsafe { COUNTER.fetch_add(1, atomic::Ordering::Relaxed) };
+            COUNTER.fetch_add(1, atomic::Ordering::Relaxed);
```
adamgreig: Should just be `Ordering::Relaxed`

src/concurrency/concurrency.md
```diff
@@ -174,30 +174,30 @@ fn main() -> ! {
...
         }
     }
 }
 
 #[interrupt]
 fn timer() {
     // Use `store` to write 0 directly to COUNTER
-    unsafe { COUNTER.store(0, atomic::Ordering::Relaxed) }
+    COUNTER.store(0, atomic::Ordering::Relaxed)
```
adamgreig: `Ordering::Relaxed`

src/concurrency/concurrency.md
japaric: fixed

PR #73, opened 2018-11-05
interop: point to the cstr-core and cty

japaric:
these are the no_std versions of std::ffi and std::os::raw

### DIFF
--- a/src/interoperability/interoperability.md
+++ b/src/interoperability/interoperability.md
@@ -16,7 +16,7 @@ is similar enough or the same.
 types such as Strings, mapping both `&str` and `String`
 to C-types that are easier and safer to handle.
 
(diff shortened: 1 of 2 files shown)


src/interoperability/interoperability.md
```diff
@@ -16,8 +16,12 @@ is similar enough or the same.
 types such as Strings, mapping both `&str` and `String`
 to C-types that are easier and safer to handle.
 
-For `#![no_std]` environments, there is also `core::raw` that
-covers all the primitive types.
+Neither of these modules in available in `core`, but you can find a `#![no_std]`
```
jamesmunns:
```suggestion
Neither of these modules are available in `core`, but you can find a `#![no_std]`
```

PR #81, opened 2018-11-13
[WIP] First draft of c-tips chapter

adamgreig:
This is still pretty rough but I wanted to get the PR open <sup><sup>before the meeting</sup></sup> and start getting feedback.

* What other topics should this chapter cover?
* What level of detail do you think is appropriate for the topics? I sort of want to just point people in the right direction and maybe give some minimal examples rather than rephrasing existing documentation. Especially I didn't want to write a general summary of things that are not really embedded (or C) specific.
* What sort of external resources should we link to? There's a lot of blog posts talking about most of these concepts but many are not really embedded (or even C) specific.

### DIFF
--- a/src/c-tips/index.md
+++ b/src/c-tips/index.md
@@ -54,6 +54,7 @@ pub mod iir;
 You can similarly include code blocks only if a feature is _not_ enabled, or if
 any combination or features is or is not enabled.
 
+The feature selection will only apply to the next statement or block. If a block can not be used in the current scope then then `cfg` attribute will need to be used multiple times.
 It's worth noting that most of the time it is better to simply include all the
 code and allow the compiler to remove dead code when optimising: it's simpler
 for you and your users, and in general the compiler will do a good job of
--- a/src/c-tips/index.md
+++ b/src/c-tips/index.md
@@ -192,7 +192,7 @@ data processing code.
 See the [Iterators in the Book] and [Iterator documentation] for more details.
 
 [Iterators in the Book]: https://doc.rust-lang.org/book/second-edition/ch13-02-iterators.html
-[Iterator method documentation]: https://doc.rust-lang.org/core/iter/trait.Iterator.html
+[Iterator documentation]: https://doc.rust-lang.org/core/iter/trait.Iterator.html
 
 ## References vs Pointers
 
--- a/src/c-tips/index.md
+++ b/src/c-tips/index.md
@@ -218,6 +218,82 @@ with hardware (for example, writing a pointer to a buffer into a DMA peripheral
 register), and they are also used under the hood for all peripheral access
 crates to allow you to read and write memory-mapped registers.
 
+## Volatile Access
+
+In C, individual variables may be marked `volatile`, indicating to the compiler
+that the value in the variable may change between accesses. Volatile variables
+are commonly used in an embedded context for memory-mapped registers.
+
+In Rust, instead of marking a variable as `volatile`, we use specific methods
+to perform volatile access: [`core::ptr::read_volatile`] and
+[`core::ptr::write_volatile`]. These methods take a `*const T` or a `*mut T`
+(_raw pointers_, as discussed above) and perform a volatile read or write.
+
+[`core::ptr::read_volatile`]: https://doc.rust-lang.org/core/ptr/fn.read_volatile.html
+[`core::ptr::write_volatile`]: https://doc.rust-lang.org/core/ptr/fn.write_volatile.html
+
+For example, in C you might write:
+
+```c
+volatile bool signalled = false;
+
+void ISR() {
+    // Signal that the interrupt has occurred
+    signalled = true;
+}
+
+void driver() {
+    while(true) {
+        // Sleep until signalled
+        while(!signalled) { WFI(); }
+        // Reset signalled indicator
+        signalled = false;
+        // Perform some task that was waiting for the interrupt
+        run_task();
+    }
+}
+```
+
+The equivalent in Rust would use volatile methods on each access:
+
+```rust
+static mut SIGNALLED: bool = false;
+
+#[interrupt]
+fn ISR() {
+    // Signal that the interrupt has occurred
+    // (In real code, you should consider a higher level primitive,
+    //  such as an atomic type).
+    unsafe { core::ptr::write_volatile(&mut SIGNALLED, true) };
+}
+
+fn driver() {
+    loop {
+        // Sleep until signalled
+        while unsafe { !core::ptr::read_volatile(&SIGNALLED) } {}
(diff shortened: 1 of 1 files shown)


adamgreig:
From last week's meeting:

* Add `repr(align)` and probably also other `repr`s to topics

src/c-tips/c-tips.md
```diff
@@ -1,3 +1,209 @@
...
+#[cfg(feature="FIR")]
+pub mod fir;
+
+#[cfg(feature="IIR")]
+pub mod iir;
+```
+
+You can similarly include code blocks only if a feature is _not_ enabled, or if
+any combination or features is or is not enabled.
+
```
therealprof: Can we add something like: 
```suggestion

The feature selection will only apply to the next statement or block. If a block can not be used in the current scope then then `cfg` attribute will need to be used multiple times.
```

src/c-tips/c-tips.md
```diff
@@ -1,3 +1,209 @@
...
+data processing code.
+
+See the [Iterators in the Book] and [Iterator documentation] for more details.
+
+[Iterators in the Book]: https://doc.rust-lang.org/book/second-edition/ch13-02-iterators.html
+[Iterator method documentation]: https://doc.rust-lang.org/core/iter/trait.Iterator.html
+
+## References vs Pointers
+
+In Rust, pointers (called [_raw pointers_]) exist but are rarely used:
```
therealprof: I wouldn't deem this statement accurate. Especially with embedded Rust raw pointer have to be used in many low level places like memory mapped registers and DMA.

src/c-tips/c-tips.md
```diff
@@ -1,3 +1,209 @@
...
+the same value at any given time.
+
+In practice this means you have to be more careful about whether you need
+mutable access to data: where in C the default is mutable and you must be
+explicit about `const`, in Rust the opposite is true.
+
+## Other Resources
+
+* In this book: [A little C with your Rust](../interoperability/c-with-rust.md)
+* In this book: [A little Rust with your C](../interoperability/rust-with-c.md)
```
therealprof: **In this book** duplicated on following lines

adamgreig: This is deliberate in so far as those two items are each in the book and the subsequent items are not. I think it would look OK rendered?

src/c-tips/c-tips.md
```diff
@@ -1,3 +1,209 @@
...
+int16_t arr[16];
+int i;
+for(i=0; i<sizeof(arr)/sizeof(arr[0]); i++) {
+    process(arr[i]);
+}
+```
+
+In Rust this is an anti-pattern: indexed access can be slower (as it needs to
+be bounds checked) and may prevent various compiler optimisations.
+
```
korken89: Something like this addition perhaps, to really highlight this very important distinction? Edit as seen fit. :)

```suggestion
It is important to see this distinction in Rust vs C, as Rust needs to check for out-of-bounds access when manually indexing an array to guarantee memory safety, while C will happily index outside the array if given an index that will do so.  
```

src/c-tips/c-tips.md
adamgreig: Good point, I've updated to indicate this.

src/c-tips/c-tips.md
```diff
@@ -1,3 +1,209 @@
...
+Common use cases for build scripts include:
+
+* provide build-time information, for example statically embedding the build
+  date or Git commit hash into your executable
+* generate linker scripts at build time depending on selected features or other
+  logic
+* change the Cargo build configuration
+* add extra static libraries to link against
+
```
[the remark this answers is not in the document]
adamgreig: Thanks, have done.

src/c-tips/c-tips.md
```diff
@@ -1,3 +1,219 @@
...
+* Compile-time selection of code blocks with `#ifdef`
+* Compile-time array sizes and computations
+* Macros to simplify common patterns (to avoid function call overhead)
+
+In Rust there is no preprocessor, and so many of these use cases are addressed
+differently. In the rest of this section we cover various alternatives to
+using the preprocessor.
+
+### Compile-Time Code Selection
```
japaric: I think this section should also (briefly) mention that you can have code conditionally compile based on the target, e.g. `cfg(target_arch)`.

adamgreig: Thanks, updated to include `target_arch`.

src/c-tips/c-tips.md
```diff
@@ -1,3 +1,332 @@
...
+    z: u16,
+}
+
+fn main() {
+    let v = Foo { x: 0, y: 0, z: 0 };
+    println!("{:p} {:p} {:p}", &v.x, &v.y, &v.z);
+}
+
+// 0x7ffecb3511d0 0x7ffecb3511d4 0x7ffecb3511d2
+// Note ordering has been changed to y, z, y to improve packing.
```
korken89: Comment error
```suggestion
// Note ordering has been changed to x, z, y to improve packing.
```

src/c-tips/c-tips.md
```diff
@@ -1,3 +1,332 @@
...
+
+Iterators provide a powerful array of functionality you would have to implement
+manually in C, such as chaining, zipping, enumerating, finding the min or max,
+summing, and more. Iterator methods can also be chained, giving very readable
+data processing code.
+
+See the [Iterators in the Book] and [Iterator documentation] for more details.
+
+[Iterators in the Book]: https://doc.rust-lang.org/book/second-edition/ch13-02-iterators.html
+[Iterator method documentation]: https://doc.rust-lang.org/core/iter/trait.Iterator.html
```
korken89: Error in link
```suggestion
[Iterator documentation]: https://doc.rust-lang.org/core/iter/trait.Iterator.html
```

adamgreig:
Thanks for spotting those errors!

adamgreig:
Volatile is a great point. I'll add it to this PR now.

adamgreig:
@jamesmunns Updated to mention volatile.

src/c-tips/c-tips.md
adamgreig: I've updated the list to nest these two under a single "In this book", which looks better. Thanks for pointing it out.

adamgreig:
I have rebased from master to fix the merge conflict.

adamgreig:
Looks like the link back to the concurrency chapter is failing; should that be a link to `../concurrency/index.md`?

jamesmunns:
@adamgreig if you're ready to merge this, I can hit it with an `r+`.

adamgreig:
I think it's ready, thanks!

Issue #65, opened 2018-10-30
Extend Portability Chapter


jamesmunns:
Moved from #36 

I would like to see the following introduced to the `embedded-hal`/portability chapter:

* Examples (we've already discussed)
* Commentary on why these traits are useful, (N*M vs N+M implementation, where N is # of chip/board support crates, and M is the # of driver traits).

Additionally, we may want something like [this image](https://screenshots.firefox.com/RviflD6enGaYiZRm/ferrous-systems.github.io), to demonstrate how this all fits together, and we may want to talk about the different components in the ecosystem somewhere:

* Peripheral Access Crates
* HAL Implementations
* Board Support Crates
* `embedded-hal`
* Driver Crates
* `no_std` libraries/crates (such as `heapless`, `menu`, etc.)

jamesmunns:
Addressed by #83 and #84, thanks @therealprof!

jamesmunns:
Still to do: Examples of each layer

TotalKrill:
After reading this book I still  feel like this chapter needs to be extended, I got a clear view of what the different crates shoul be (hal, PAC, drivers). But i do not feel that I have any clue on how to start utilizing these, or how to start creating or using drivers for embedded-hal, nor do i get any practical understanding of the benefits of rust ecosystem here. I actually felt more optimistic towards portability due to driver crates and embedded hal before reading this. 

I also feel I have no clear view of how to structure a project where I have to develop an embedded system using rust, and there are new devices that I need to fix a driver for.

Maybe some example on how to create a new embedded hal-crate for something existing on the stm32f3discovery, such as the acceleremoter, maybe a walk-through an example of how to get an embedded-hal trait object such as i2c from the discovery hal crate, create a basic driver for the accelerometer that accepts the i2c traits and is therefore portable. 

to tie everything together in the end, this could be extended to react on an interrupt event of the accelerometer, thus tying into the benefits of rust concurrancy etc.

korken89:
Ping @rust-embedded/resources

PR #83, opened 2018-11-14
Mention complexity benefits by using embedded-hal

therealprof:


therealprof:
CC: #65

src/portability/index.md
```diff
@@ -28,6 +28,8 @@ Some of the defined traits in **embedded-hal** are:
 * Timers/Countdowns
 * Analog Digital Conversion
 
+The main reason for having the **embedded-hal** traits and crates implementing and using them is to keep complexity in check. If you consider that an application might have to implement the use of the peripheral in the hardware as well as the application and potentially drivers for additional hardware components, then it should be easy to see that the reusablility is very limited. Expressed mathematically, if **M** is the number of peripheral HAL implementations and **N** the number of drivers then if we were to reinvent the wheel for every application then we would end up with **M*N** implementations while by sticking to the *API* provided by the **embedded-hal** traits will reduce the implementation complexity to **M+N**. Of course there're additional benefits to be had like less trial-and-error due to a well-defined and ready-to-use APIs.
```
korken89:
```suggestion
The main reason for having the **embedded-hal** traits and crates implementing and using them is to keep complexity in check. If you consider that an application might have to implement the use of the peripheral in the hardware as well as the application and potentially drivers for additional hardware components, then it should be easy to see that the re-usability is very limited. Expressed mathematically, if **M** is the number of peripheral HAL implementations and **N** the number of drivers then if we were to reinvent the wheel for every application then we would end up with **M*N** implementations while by using the *API* provided by the **embedded-hal** traits will make the implementation complexity approach **M+N**. Of course there're additional benefits to be had, such as less trial-and-error due to a well-defined and ready-to-use APIs.
```

therealprof:
@korken89 Thanks.

PR #84, opened 2018-11-14
Added layer diagram for embedded-hal use

therealprof:
Signed-off-by: Daniel Egger <daniel@eggers-club.de>


therealprof:
@jamesmunns How about this?

therealprof:
CC: #65

jamesmunns:
@therealprof a couple thoughts, mostly about the diagram:

1. Do you want to include `board support crates` in the stack?
2. Additionally with `drivers`, you might want to include the difference between `peripheral drivers` and `external hardware drivers`, as I think peripheral drivers will be below the HAL line, and ext HW drivers are above the HAL line
3. Do you want to mention libraries/crates that are not specifically hardware related? For example something from [the no_std crate list](https://github.com/rust-embedded/awesome-embedded-rust#no-std-crates), like `menu` or `smoltcp`?
    * I see these as below the application, but above the embedded-hal line (usually)

therealprof:
> Do you want to include board support crates in the stack?

Sure, will do.

> Additionally with drivers, you might want to include the difference between peripheral drivers and external hardware drivers, as I think peripheral drivers will be below the HAL line, and ext HW drivers are above the HAL line

The situation here is a bit more tricky. We have drivers on both sides but also drivers which basically use a HAL impl and provide another HAL impl (like I2C multiplexers). Not sure how to express that.

> Do you want to mention libraries/crates that are not specifically hardware related? For example something from the no_std crate list, like menu or smoltcp?
I see these as below the application, but above the embedded-hal line (usually)

No, not really. An application will usually use additional crates which are completely independent of what we're trying to express here.

Also I think we should be careful not to overload the picture since there're a lot more variants and we cannot really show them all. I think the diagram should convey the basic idea and not suggest it is required to be done that way.

therealprof:
@jamesmunns @thejpster Updated diagram. Let me know what you think.

thejpster:
LGTM

korken89:
Looking good, will you approve and bors it if you are happy with it @jamesmunns ?

jamesmunns:
Yup, looks good!
bors r+

PR #89, opened 2018-11-19
Few typo fixes

adamgreen:
I fixed a few more typos I noticed as I finished my first read through of the Embedded Rust Book.

My editor also deleted some trailing white space from a few lines when I saved out my typo fixes.


src/portability/index.md
```diff
@@ -53,7 +53,7 @@ Such a **HAL implementation** can come in various flavours:
 
 ### Driver
 
-A driver implements a set of custom functionality for an internal or external component, connected to a peripheral implementing the embedded-hal traits. Typical examples for such drivers include various sensors (temperatur, magnetometer, accelerometer, light), display devices (LED arays, LCD displays) and actors (motors, relais).
+A driver implements a set of custom functionality for an internal or external component, connected to a peripheral implementing the embedded-hal traits. Typical examples for such drivers include various sensors (temperature, magnetometer, accelerometer, light), display devices (LED arrays, LCD displays) and actuators (motors, relays).
```
therealprof: Hm, for the specified examples `actuator` would fit but I really meant to say `actor` which is a commonly used term as counterpart of sensor. Maybe replace the "relays" by "transmitter" to ensure that is does not only cover moving parts?

adamgreen: Ok, my mistake, I will make those changes. I have almost never seen the term actors used in that way. Maybe it's an American/Queen's English difference?

therealprof: I don't know. And reading up on the topic I actually think that actor is also probably not the best term to use here as it seems to refer to the system as a whole and not individual components. That sentence would probably use a good overhaul, either by finding a more suitable category for both moving and non-moving parts or by splitting it up.

PR #90, opened 2018-11-19
Move "last_state = state" assignment

adamgreen:
I believe that I have found a bug in the samples found in the concurrency chapter where the value of the `last_state` variable should be updated on each loop iteration but it is only updated on the detection of the first leading edge. For example:

```rust
    loop {
        let state = read_signal_level();
        if state && !last_state {
            last_state = state;
            // DANGER - Not actually safe! Could cause data races.
            unsafe { COUNTER += 1 };
        }
    }
```

I think that loop should actually be written as:
```rust
    loop {
        let state = read_signal_level();
        if state && !last_state {
            // DANGER - Not actually safe! Could cause data races.
            unsafe { COUNTER += 1 };
        }
        last_state = state;
    }
```

As the code exists today, I think it would detect the first leading edge, set last_state to true and then never be updated again which would also mean that it never detects another leading edge.

### DIFF
--- a/src/concurrency/index.md
+++ b/src/concurrency/index.md
@@ -68,10 +68,10 @@ fn main() -> ! {
     loop {
         let state = read_signal_level();
         if state && !last_state {
-            last_state = state;
             // DANGER - Not actually safe! Could cause data races.
             unsafe { COUNTER += 1 };
         }
+        last_state = state;
     }
 }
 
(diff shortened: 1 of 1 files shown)


therealprof:
This example is rather weird anyway: I would not expect a function called `read_signal_level()` to return a `bool` and in the case of a non-bool the whole example falls flat on its face.

therealprof:
@adamgreen Understood, but I still find the naming of the fictive function to be a potential source of confusion, especially since the type is inferred.

PR #106, opened 2018-12-17
Added section about interrupts, closes #98

### LINKED ISSUE #98: document #[interrupt]
svd2rust 0.14.0 is out so this is now unblocked

therealprof:
Signed-off-by: Daniel Egger <daniel@eggers-club.de>

### DIFF
--- a/src/start/interrupts.md
+++ b/src/start/interrupts.md
@@ -1,0 +1,54 @@
+# Interrupts
+
+Interrupts differ from exceptions in a variety of ways but their operation and
+use is largely similar and they are also handled by the same interrupt
+controller. Whereas exeptions are defined by the Cortex-M architecture,
+interrupts are always vendor (and often even chip) specific implementations,
+both in naming and functionality.
+
+Interrupts do allow for a lot of flexbility which needs to be accounted for
+when attempting to use them in an advanced way. We will not cover those uses in
+this book, however it is a good idea to keep the following in mind:
+
+* Interrupts have programmable priorities which determine their handlers' execution order
+* Interrupts can nest and preempt, i.e. execution of an interrupt handler might be interrupted by another higher-priority interrupt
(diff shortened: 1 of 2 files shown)


src/start/interrupts.md
```diff
@@ -0,0 +1,54 @@
...
+
+Similar to exeption handlers it is also possible to declare `static mut`
+variables inside the interrupt handlers for *safe* state keeping.
+
+``` rust,ignore
+#[interrupt]
+fn TIM2() {
+    static mut COUNT: u32 = 0;
+
+    // `COUNT` has type `&mut u32` and it's safe to use
```
korken89: This is in contradiction with the definition, as the interrupt macro does this conversion for us.
I would expand this comment or discuss to not have readers be confused.

src/start/interrupts.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+Interrupt handlers look like plain functions (except for the lack of arguments)
+similar to exception handlers. However they can not be called directly by other
+parts of the firmware due to the special calling conventions. It is however
+possible to generate interrupt requests in software to trigger a diversion to
+to the interrupt handler.
+
+Similar to exeption handlers it is also possible to declare `static mut`
+variables inside the interrupt handlers for *safe* state keeping.
```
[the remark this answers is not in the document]
therealprof: @korken89 That explanation is in the exception section already and I didn't want to copy half of the section in verbatim. Maybe add an anchor and a link?

src/start/interrupts.md
therealprof: Again, that is taken from the exception section. ;)

korken89: I think a reference will do perfectly! Just to minimize the chance of misunderstanding :)

therealprof:
@korken89 Added the link.

Issue #123, opened 2019-02-06
Fix Contributors List


jamesmunns:
https://github.com/rust-embedded/book/blob/master/book.toml#L2 is wildly wrong, and I think it makes sense to add a page of contributors to the book.

At least, we should point back to the resources team, but I would be in favor of having a "Thanks" page at the end of the book, listing all of the contributors.

PR #124, opened 2019-02-06
Add basic contributor's page based on git commit history

Nicoretti:
* Add inital contributor's page
* Add script to generate conributors page

Related Issues:
* https://github.com/rust-embedded/book/issues/123

### DIFF
--- a/ci/contributors.py
+++ b/ci/contributors.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+import os
+import sys
+import subprocess
+from pathlib import Path
+from functools import partial
+from collections import namedtuple, defaultdict
+
+Committer = namedtuple('Committer', ['name', 'email'])
+
+BLACKLIST = ['GitHub', 'bors[bot]']
+OVERWRITES = {
+        'James Munns': {'name': 'James Munns', 'email': 'james.munns@ferrous-systems.com'},
+}
+
+
+def committer_log():
+    """
+    Generator, returning the commiter for each commit in the git repository.
+    """
+    command = ['git', 'log', '--pretty=%cn;%ce']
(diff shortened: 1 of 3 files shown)


Nicoretti:
@jamesmunns I've seen the https://github.com/rust-embedded/book/issues/123 you created, this PR might be a first start to address this issue. The script `contributors.sh` could be extended and hooked up with the ci.

greez
Nico

therealprof:
Hm, there're some duplicates in the contributors list. Also some people might prefer to be mentioned by their nick instead of their specified real name. We might have to do some postprocessing here, otherwise looks good to me.

Nicoretti:
@therealprof sure, thing I was anyway meant a draft to discuss up on. I'm could think of adding mappings for e.g. the nicknames. However if it shall be more complex I would prefer to do it 
e.g. in python instead of using a shell script.

therealprof:
Using Python should be fine.

jamesmunns:
Hey @Nicoretti, thank you for the contribution! I'll do a quick vote of @rust-embedded/resources to see if we want to recognize people in this way.

If so, we can figure out some of the details around duplication, automation, etc.

- [ ] @adamgreig 
- [ ] @ryankurte 
- [ ] @korken89 
- [ ] @thejpster 
- [ ] @jamesmunns 
- [ ] @therealprof

jamesmunns:
CC @japaric - this would be another place where "bors runs some kind of script on every merge to master" would be useful.

adamgreig:
Given the issues around duplication, people potentially not wanting their email address linked to, and not being able to have bors run a script yet, I think my vote would be to not have the Python script in the repo, but to run it once to generate a list of names, and then manually maintain the list and suggest new contributors add themselves to it if they want to when they submit PRs.

I'd probably have either a list of names or names linked to GitHub profiles, rather than email addresses.

Issue #160, opened 2019-02-17
Jitter ?


flip111:
https://rust-embedded.github.io/book/concurrency/index.html

> which comes at a cost to overhead and interrupt latency and jitter.

I don't really understand what the consequences are from disabled interrupts. Is it not so that interrupts will just be ignored ? But latency indicates they will be delayed? And what about jitter?

adamgreig:
I think explaining interrupt latency and jitter is out of scope for the book (and especially for the concurrency section); the introduction says we assume familiarity with interrupts in an embedded context.

> Is it not so that interrupts will just be ignored ? But latency indicates they will be delayed? And what about jitter?

When interrupt handling is disabled in a critical section, interrupts are left pending, and will be serviced once interrupt handling is re-enabled (hence the additional latency and jitter).

PR #164, opened 2019-02-17
Update index.md

flip111:
https://github.com/rust-embedded/book/issues/160


adamgreig:
My preference is to not change this, but if we do I'd rather reword it to flow better. Something like:

> This solves our immediate problem, but we're still left writing a lot of `unsafe` code which we need to carefully reason about, and we might be using critical sections needlessly. Since each critical section temporarily pauses interrupt processing, there is an associated cost of some extra code size and higher interrupt latency and jitter (interrupts may take longer to be processed, and the time until they are processed will be more variable). Whether this is a problem depends on your system, but in general we'd like to avoid it.

PR #165, opened 2019-02-17
Clarify list of available targets for installation

adamgreig:


Disasm:
Ping from triage: @korken89, could you review this?

Issue #134, opened 2019-02-17
why mention git ?


flip111:
Don't understand why this heading mention `git`

https://github.com/rust-embedded/book/blob/01327046e391d59de5de0079df30d1b57533e9da/src/intro/tooling.md#cargo-generate-or-git

only reference i see that it can be used as a download tools. But since there are more download tools they could all be put in the heading as well.

please clarify

flip111:
So if anyone stumbled on this. `cargo generate`, git, curl, wget or your browser are all good. Only two tools were mentioned  to keep the title small.

Issue #143, opened 2019-02-17
wording way too relaxed about not installing cargo generate


flip111:
https://rust-embedded.github.io/book/intro/tooling.html

> [OPTIONAL] git OR cargo-generate. If you have neither installed then don't worry about installing either.

Big fat `OPTIONAL` and `don't worry`. Later https://rust-embedded.github.io/book/start/qemu.html#a-non-standard-rust-program it's required to follow the book because at that point no alternative is provided.

Suggestion 1:
> cargo-generate and git. This is optional for embedded programming with rust but needed to follow along with the book.

Suggestion 2:
provide install command for cargo-generate here https://rust-embedded.github.io/book/intro/install.html also point to git website

PR #166, opened 2019-02-18
Update tooling.md

flip111:
https://github.com/rust-embedded/book/issues/134

### DIFF
--- a/src/intro/tooling.md
+++ b/src/intro/tooling.md
@@ -15,15 +15,13 @@ tested.
 - OpenOCD >=0.8. Tested versions: v0.9.0 and v0.10.0
 - GDB with ARM support. Version 7.12 or newer highly recommended. Tested
   versions: 7.10, 7.11, 7.12 and 8.1
-- [OPTIONAL] `git` OR
-  [`cargo-generate`](https://github.com/ashleygwilliams/cargo-generate). If you
-  have neither installed then don't worry about installing either.
+- [`cargo-generate`](https://github.com/ashleygwilliams/cargo-generate) and `git`.
+These tools are optional but will make it easier to follow along with the book.
 
 The text below explains why we are using these tools. Installation instructions
 can be found on the next page.
 
-## `cargo-generate` OR `git`
-
+## Using a template for the first project ##
 Bare metal programs are non-standard (`no_std`) Rust programs that require some
 fiddling with the linking process to get the memory layout of the program
 right. All this requires unusual files (like linker scripts) and unusual


flip111:
https://github.com/rust-embedded/book/issues/143

src/intro/tooling.md
```diff
@@ -15,15 +15,13 @@ tested.
...
-  have neither installed then don't worry about installing either.
+- [`cargo-generate`](https://github.com/ashleygwilliams/cargo-generate) and `git`.
+These tools are optional but will make it easier to follow along with the book.
 
 The text below explains why we are using these tools. Installation instructions
 can be found on the next page.
 
-## `cargo-generate` OR `git`
-
+## Using a template for the first project ##
```
japaric: All headings are named after tools. With this change that would no longer be the case. I would prefer to keep the symmetry.

src/intro/tooling.md
```diff
@@ -15,15 +15,13 @@ tested.
 - OpenOCD >=0.8. Tested versions: v0.9.0 and v0.10.0
 - GDB with ARM support. Version 7.12 or newer highly recommended. Tested
   versions: 7.10, 7.11, 7.12 and 8.1
-- [OPTIONAL] `git` OR
-  [`cargo-generate`](https://github.com/ashleygwilliams/cargo-generate). If you
-  have neither installed then don't worry about installing either.
+- [`cargo-generate`](https://github.com/ashleygwilliams/cargo-generate) and `git`.
```
japaric: This seems fine to me

src/intro/tooling.md
flip111: What about `cargo-generate and other tools` ?

src/intro/tooling.md
japaric: Actually, this should be "(..) or `git`". You don't need *both* tools; one of them suffices.

src/intro/tooling.md
flip111: By the way, maybe write something about rustc too ..

japaric: The current text matches what's listed in the bullet list. The "other tools" in your proposal makes it unclear whether "other tools" means just git or all / some of the other tools in the bullet list.

Personally, I think the current text is fine.

src/intro/tooling.md
flip111: Later on in the book `cargo generate --git` command is used. We can put `or` here when comes with it's own git api and doesn't rely on standalone git. I assumed it relies on standalone git.

flip111: From this wording i conclude that standalone git is not required https://github.com/rust-lang/git2-rs#version-of-libgit2 (dependency of cargo generate)

src/intro/tooling.md
japaric: > By the way, maybe write something about rustc too ..

Rust developers (the target audience of this book) are familiar with rustc / Cargo; I don't think it's necessary to add a section for them.

src/intro/tooling.md
japaric: `cargo-generate` doesn't shell out to `git`; it uses libgit.

src/intro/tooling.md
flip111: i just find it strange that the choice now is between `cargo generate` and `git` both tools need to be downloaded and installed, while the other tools "curl, wget, or your web browser." are likely to be already on a system. But i'm tired to bike shed over this, so you can get the final call, i'll leave it as it is or make it whatever.

japaric:
@flip111, in that case let's leave the text as it is. Could you please close the two issues you linked from here? Thanks.

flip111:
I was referring to line 24 which you commented on. Since this PR is now closed i opened a new PR with only the pieces that didn't get negative feedback. https://github.com/rust-embedded/book/pull/172 I will close the second issue later

Issue #139, opened 2019-02-17
let's get familiar with hardware should have been put earlier


flip111:
https://rust-embedded.github.io/book/intro/hardware.html

But in section before that, already diving into the hardware

https://rust-embedded.github.io/book/intro/install/verify.html

Also in the installation section it would be handy to already know (which is discussed in the hardware chapter) which CPU the board has. Because then it can be verified the right toolchain is being installed for the CPU model.

Suggestion:
Move chapter 1.4 between 1.2 and 1.3

PR #168, opened 2019-02-18
Update SUMMARY.md

flip111:


flip111:
https://github.com/rust-embedded/book/issues/139

andre-richter:
Makes sense to me. Even in the tooling section the hardware board is referenced by exact name.

bors r+

Issue #140, opened 2019-02-17
`IMPORTANT` section of 3.3 volt ambigious


flip111:
https://rust-embedded.github.io/book/intro/hardware.html

It's strange that `pins` and `headers` are quoted. Assuming a really really hardware nooby here which doesn't know what a pin is, so it has to be quoted to clarify it's a special term.

Then to continue to say IMPORTANT the board works at 3.3 volt without any clarification why that is important or what they user should do.

Suggestion 1:
Think about hardware level knowledge of reader, adjust text accordingly. Not in one section assume a lot of knowledge and in another part assume no knowledge.

Suggestion 2:
Clarify 3.3 volt warning or remove it.

Suggestion 3:
Point to some manufacturer safety document or to the general documentation.

PR #169, opened 2019-02-18
Update hardware.md

flip111:
https://github.com/rust-embedded/book/issues/140

### DIFF
--- a/src/intro/hardware.md
+++ b/src/intro/hardware.md
@@ -36,4 +36,4 @@ What does this board contain?
 
 For a more detailed features list and further specification of the board take a look at [STMicroelectronics](https://www.st.com/en/evaluation-tools/stm32f3discovery.html) website.
 
-A word of warning: be careful if you want to apply external signals to the board. The microcontroller STM32F303VCT6 pins take a maximum voltage of 4 volts, but ideally should be driven by the nominal voltage of 3.3 volts. For further information consult the [6.2 Absolute maximum ratings section in the manual](https://www.st.com/resource/en/datasheet/stm32f303vc.pdf)
+A word of warning: be careful if you want to apply external signals to the board. The microcontroller STM32F303VCT6 pins take a nominal voltage of 3.3 volts. For further information consult the [6.2 Absolute maximum ratings section in the manual](https://www.st.com/resource/en/datasheet/stm32f303vc.pdf)
--- a/src/intro/hardware.md
+++ b/src/intro/hardware.md
@@ -8,11 +8,9 @@ Let's get familiar with the hardware we'll be working with.
 <img title="F3" src="../assets/f3.jpg">
 </p>
 
-We'll refer to this board as "F3" throughout this book.
-
 What does this board contain?
 
-- A STM32F303VCT6 microcontroller. This microcontroller has
+- A [STM32F303VCT6](https://www.st.com/en/microcontrollers/stm32f303vc.html) microcontroller. This microcontroller has
   - A single-core ARM Cortex-M4F processor with hardware support for single-precision floating point
     operations and a maximum clock frequency of 72 MHz.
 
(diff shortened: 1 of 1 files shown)


src/intro/hardware.md
```diff
@@ -8,37 +8,32 @@ Let's get familiar with the hardware we'll be working with.
...
-- 8 user LEDs arranged in the shape of a compass
+- A second microcontroller: a [STM32F103CBT](https://www.st.com/en/microcontrollers/stm32f103cb.html). This microcontroller is actually part of an on-board programmer and debugger named ST-LINK and is connected to the USB port named "USB ST-LINK".
 
-- A second microcontroller: a STM32F103CBT. This microcontroller is actually part of an on-board
-  programmer and debugger named ST-LINK and is connected to the USB port named "USB ST-LINK".
+For a more detailed features list and further specification of the board take a look at [STMicroelectronics](https://www.st.com/en/evaluation-tools/stm32f3discovery.html) website.
 
-- There's a second USB port, labeled "USB USER" that is connected to the main microcontroller, the
-  STM32F303VCT6, and can be used in applications.
+A word of warning: be careful if you want to apply external signals to the board. The microcontroller STM32F303VCT6 pins take a maximum voltage of 4 volts, but ideally should be driven by the nominal voltage of 3.3 volts. For further information consult the [6.2 Absolute maximum ratings section in the manual](https://www.st.com/resource/en/datasheet/stm32f303vc.pdf)
```
therealprof: I would not mention "4V" (not even sure whether that's a typo or the maximum tolerable regular 3.3V) but instead say the nominal voltage is 3.3V with only a few pins being 5V tolerant.

src/intro/hardware.md
```diff
@@ -8,37 +8,32 @@ Let's get familiar with the hardware we'll be working with.
...
+- A [magnetometer](https://en.wikipedia.org/wiki/Magnetometer) as part of the [LSM303DLHC](https://www.st.com/en/mems-and-sensors/lsm303dlhc.html) chip.
 
-- A [gyroscope].
+- A [gyroscope](https://en.wikipedia.org/wiki/Gyroscope) as part of the [L3GD20](https://www.pololu.com/file/0J563/L3GD20.pdf) chip.
 
-[gyroscope]: https://en.wikipedia.org/wiki/Gyroscope
+- 8 user LEDs arranged in the shape of a compass.
 
-- 8 user LEDs arranged in the shape of a compass
+- A second microcontroller: a [STM32F103CBT](https://www.st.com/en/microcontrollers/stm32f103cb.html). This microcontroller is actually part of an on-board programmer and debugger named ST-LINK and is connected to the USB port named "USB ST-LINK".
```
therealprof: While it is possible that there's a new DISC1 version of this board, in many cases the ST-Link/v2 is implemented via a STM32F103C8 MCU. I'd prefer to leave out the the part after the 103 to cater for both cases.

src/intro/hardware.md
flip111: I agree with this.

src/intro/hardware.md
flip111: Not sure what you are getting at. Is there an alternative version of the same board? Shouldn't it have a different part number then?

therealprof: > Not sure what you are getting at. Is there an alternative version of the same board? Shouldn't it have a different part number then?

Yes. Typically the discoveries have a different part number but in this case there're just three revisions: "On-board ST-LINK/V2 for PCB version A or B or ST-LINK/V2-B for PCB version C and newer". Revision A and B have a STM32F103C8, revision C is a STM32F103CB.

flip111: What about

> This microcontroller is actually part of an on-board programmer / debugger and is connected to the USB port named "USB ST-LINK".

therealprof: You mean striking the previous sentence? Works for me.

src/intro/hardware.md
```diff
@@ -8,37 +8,32 @@ Let's get familiar with the hardware we'll be working with.
...
+- A [magnetometer](https://en.wikipedia.org/wiki/Magnetometer) as part of the [LSM303DLHC](https://www.st.com/en/mems-and-sensors/lsm303dlhc.html) chip.
 
-- A [gyroscope].
+- A [gyroscope](https://en.wikipedia.org/wiki/Gyroscope) as part of the [L3GD20](https://www.pololu.com/file/0J563/L3GD20.pdf) chip.
 
-[gyroscope]: https://en.wikipedia.org/wiki/Gyroscope
+- 8 user LEDs arranged in the shape of a compass.
 
-- 8 user LEDs arranged in the shape of a compass
+- A second microcontroller: a [STM32F103CBT](https://www.st.com/en/microcontrollers/stm32f103cb.html). This microcontroller is actually part of an on-board programmer / debugger and is connected to the USB port named "USB ST-LINK".
```
therealprof: The STM32F103CBT reference is the one that required the fixing...

flip111: Can you try to edit my patch? is that enabled?

therealprof:
```suggestion
- A second microcontroller: a [STM32F103](https://www.st.com/en/microcontrollers/stm32f103cb.html). This microcontroller is actually part of an on-board programmer / debugger and is connected to the USB port named "USB ST-LINK".
```

PR #172, opened 2019-02-18
Update tooling.md

flip111:
https://github.com/rust-embedded/book/issues/143


andre-richter:
r? @japaric 

This is directly related to your review in  #166. I think it makes sense you therefore conclude this.

PR #178, opened 2019-03-10
Provide a hint for installing cargo-generate

nvll:
I wasn't familiar with `cargo-generate` so I had to google and find Ashley's repo explaining it. Since this is catering to new embedded users but not necessarily new rust users it may not be a useful suggestion depending on POV.

### DIFF
--- a/src/start/qemu.md
+++ b/src/start/qemu.md
@@ -14,7 +14,7 @@ project from it.
 
 [`cortex-m-quickstart`]: https://github.com/rust-embedded/cortex-m-quickstart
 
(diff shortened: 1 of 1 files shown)


jamesmunns:
Hey @nvll! This looks good, do you think it would make more sense to have this in the `install` section of the book instead? I'm not sure if we show installation instructions anywhere else in the book at the moment.

nvll:
Sure, I'll take a look a that section and find a place for it.

nvll:
Closing this to make a larger docs PR

PR #196, opened 2019-07-09
concurrency: update shared resource examples

geomatsi:
Hi all,

This PR suggests the following updates for the section on concurrency:

1. Use lazy_static macro to declare shared resources
    Suggested example of shared resource declaration does not compile
    neither on stable (edition 2018) nor on nightly with the following error:
    "error[E0015]: calls in statics are limited to constant functions, tuple structs and tuple variants"
2. Add example of mutable references to shared resources
  
Regards,
Sergey

### DIFF
--- a/src/concurrency/index.md
+++ b/src/concurrency/index.md
@@ -555,6 +555,23 @@ fn timer() {
 
 ```
 
+> **NOTE**
+>
+> At the moment `cortex-m` crate hides const versions of some functions
+> (including `Mutex::new()`) behind the `const-fn` feature.  So you need to add
+> the `const-fn` feature to dependency on cortex-m in Cargo.toml to make
+> the above examples work:
+>
+> ``` toml
+> [dependencies.cortex-m]
+> version="0.6.0"
+> features=["const-fn"]
+> ```
+> Meanwhile `const-fn` has been working on stable Rust for some time now.
+> So this additional switch in Cargo.toml will not be needed as soon as 
+> it is enabled in `cortex-m` by defauilt.
+>
+
 Whew! This is safe, but it is also a little unwieldy. Is there anything else
 we can do?
 
--- a/src/concurrency/index.md
+++ b/src/concurrency/index.md
@@ -512,8 +512,8 @@ Since we can't move the `GPIOA` out of the `&Option`, we need to convert it to
 an `&Option<&GPIOA>` with `as_ref()`, which we can finally `unwrap()` to obtain
 the `&GPIOA` which lets us modify the peripheral.
 
-If we need a mutable reference to shared resource, then `borrow_mut` and `deref_mut`
-should be used instead. The following example makes use of mutable reference TIM2 timer.
+If we need a mutable references to shared resources, then `borrow_mut` and `deref_mut`
+should be used instead. The following code shows an example using the TIM2 timer.
 
 ```rust,ignore
 use core::cell::RefCell;
@@ -531,7 +531,7 @@ fn main() -> ! {
     let dp = stm32f405::Peripherals::take().unwrap();
 
     // Some sort of timer configuration function.
-    // Assume it configures TIM2 timer, its NVIC interrupt,
+    // Assume it configures the TIM2 timer, its NVIC interrupt,
     // and finally starts the timer.
     let tim = configure_timer_interrupt(&mut cp, dp);
 
@@ -557,9 +557,9 @@ fn timer() {
 
 > **NOTE**
 >
-> At the moment `cortex-m` crate hides const versions of some functions
-> (including `Mutex::new()`) behind the `const-fn` feature.  So you need to add
-> the `const-fn` feature to dependency on cortex-m in Cargo.toml to make
+> At the moment, the `cortex-m` crate hides const versions of some functions
+> (including `Mutex::new()`) behind the `const-fn` feature. So you need to add
+> the `const-fn` feature as a dependency for cortex-m in the Cargo.toml to make
 > the above examples work:
 >
 > ``` toml
@@ -567,9 +567,9 @@ fn timer() {
 > version="0.6.0"
 > features=["const-fn"]
 > ```
-> Meanwhile `const-fn` has been working on stable Rust for some time now.
+> Meanwhile, `const-fn` has been working on stable Rust for some time now.
 > So this additional switch in Cargo.toml will not be needed as soon as 
-> it is enabled in `cortex-m` by defauilt.
+> it is enabled in `cortex-m` by default.
 >
 
 Whew! This is safe, but it is also a little unwieldy. Is there anything else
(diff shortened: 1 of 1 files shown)


adamgreig:
Thanks for the PR!

The examples of using `borrow_mut()` and `ref_mut()` are helpful, but perhaps they could be simplified to not use `ITM` at all. I think keeping the examples as simple as possible helps make them easier to understand.

I'm not a fan of adding the `lazy_static` dependency; it's not a standard part of Rust, and it adds overhead without really solving the issue, so I don't think it's a good fit for this sort of problem. The compiler error you're getting is because we need the `const-fn` feature of the `cortex-m` crate, which then exposes `Mutex::new()` as a `const fn`, allowing it to be called in this context on stable Rust. Ideally we'd make this the default on `cortex-m` but I guess that might require bumping the MSRV.

geomatsi:
Hi @adamgreen 
Thanks for review ! Ok, I will remove lazy_static part and simplify mutable shared resource example as suggested. 

By the way, could you please clarify how to fix the build for  unmodified examples ? Switching to nightly does not help. Is it necessary to explicitly enable `const-fn` at the moment ?

Regards,
Sergey

geomatsi:
BTW, dropping lazy_static means dropping the first commit of this PR. Is it ok if I force-push to this PR ?

adamgreig:
> By the way, could you please clarify how to fix the build for unmodified examples ? Switching to nightly does not help. Is it necessary to explicitly enable const-fn at the moment ?

Yep, you need to add the `const-fn` feature to your dependency on `cortex-m` in `Cargo.toml`:

```toml
cortex-m = {version="0.6.0", features=["const-fn"]}
```

Ideally we'd fix this with a new cortex-m version but it might be helpful to mention it in the book for now...

> BTW, dropping lazy_static means dropping the first commit of this PR. Is it ok if I force-push to this PR ?

Sure, no problem.

andre-richter:
Agree on dropping the lazy_static part. 👍 

There's too much magic happening behind closed curtains (IIRC it has a mutex dependency itself, has a once() function, and so on) that prevent user from really understanding whats going on, but this should be the goal of this book.

geomatsi:
Hi all,
All the requested changes have been introduced:
* remove lazy_static dependency
* simplify 'mutable shared resource' example: keep only TIM2, no need for ITM
* add note regarding `const-fn` feature for `cortex-m` crate

Regards,
Sergey

src/concurrency/index.md
```diff
@@ -512,6 +512,66 @@ Since we can't move the `GPIOA` out of the `&Option`, we need to convert it to
 an `&Option<&GPIOA>` with `as_ref()`, which we can finally `unwrap()` to obtain
 the `&GPIOA` which lets us modify the peripheral.
 
+If we need a mutable reference to shared resource, then `borrow_mut` and `deref_mut`
```
andre-richter:
```suggestion
If we need a mutable reference to shared resources, then `borrow_mut` and `deref_mut`
```

src/concurrency/index.md
```diff
@@ -512,6 +512,66 @@ Since we can't move the `GPIOA` out of the `&Option`, we need to convert it to
...
+should be used instead. The following example makes use of mutable reference TIM2 timer.
```
andre-richter:
```suggestion
should be used instead. The following code shows an example using the TIM2 timer.
```

src/concurrency/index.md
```diff
@@ -512,6 +512,66 @@ Since we can't move the `GPIOA` out of the `&Option`, we need to convert it to
...
+static G_TIM: Mutex<RefCell<Option<Timer<stm32::TIM2>>>> =
+	Mutex::new(RefCell::new(None));
+
+#[entry]
+fn main() -> ! {
+    let mut cp = cm::Peripherals::take().unwrap();
+    let dp = stm32f405::Peripherals::take().unwrap();
+
+    // Some sort of timer configuration function.
+    // Assume it configures TIM2 timer, its NVIC interrupt,
```
andre-richter:
```suggestion
    // Assume it configures the TIM2 timer, its NVIC interrupt,
```

src/concurrency/index.md
```diff
@@ -512,6 +512,66 @@ Since we can't move the `GPIOA` out of the `&Option`, we need to convert it to
...
+            tim.start(1.hz());
+        }
+    });
+}
+
+```
+
+> **NOTE**
+>
+> At the moment `cortex-m` crate hides const versions of some functions
```
andre-richter:
```suggestion
> At the moment, the `cortex-m` crate hides const versions of some functions
```

src/concurrency/index.md
```diff
@@ -512,6 +512,66 @@ Since we can't move the `GPIOA` out of the `&Option`, we need to convert it to
...
+> (including `Mutex::new()`) behind the `const-fn` feature.  So you need to add
```
andre-richter:
```suggestion
> (including `Mutex::new()`) behind the `const-fn` feature. So you need to add
```

src/concurrency/index.md
```diff
@@ -512,6 +512,66 @@ Since we can't move the `GPIOA` out of the `&Option`, we need to convert it to
...
+> the `const-fn` feature to dependency on cortex-m in Cargo.toml to make
```
andre-richter:
```suggestion
> the `const-fn` feature as a dependency for cortex-m in the Cargo.toml to make
```

src/concurrency/index.md
```diff
@@ -512,6 +512,66 @@ Since we can't move the `GPIOA` out of the `&Option`, we need to convert it to
...
+> the above examples work:
+>
+> ``` toml
+> [dependencies.cortex-m]
+> version="0.6.0"
+> features=["const-fn"]
+> ```
+> Meanwhile `const-fn` has been working on stable Rust for some time now.
```
andre-richter:
```suggestion
> Meanwhile, `const-fn` has been working on stable Rust for some time now.
```

src/concurrency/index.md
```diff
@@ -512,6 +512,66 @@ Since we can't move the `GPIOA` out of the `&Option`, we need to convert it to
...
+> So this additional switch in Cargo.toml will not be needed as soon as 
+> it is enabled in `cortex-m` by defauilt.
```
andre-richter:
```suggestion
> it is enabled in `cortex-m` by default.
```

geomatsi:
Hi @andre-richter 

Thanks a lot for your review proof-reading! I applied all the suggested changes. However I did it using github web UI adding 8 more commits. Let me know if that was a bad idea. Then I will collect them into a single commit and force-push PR.

Regards,
Sergey

andre-richter:
We should be able to squash it before merge, no worries.

Clicking the right buttons would have allowed you to accept all suggestions at one, I think you need to change to review view or something. I don't remember right now.

I'll leave it to @adamgreig to make the final call.

andre-richter:
@geomatsi Apparently, contrary to my statement, bors cannot squash yet.
So if your offer to squash and force push still holds, kindly do so :)

geomatsi:
Hi @andre-richter 

Sure, will do so later today.

Regards,
Sergey

geomatsi:
Hi @andre-richter , @adamgreig   and all,

All fixups have been merged into a single commit. Bors looks happy.

Regards,
Sergey

andre-richter:
bors r+

Thank you for your contribution :+1:

Issue #59, opened 2018-10-20
Reminder: Back out hardcoded mdbook v0.2.1 version


therealprof:
In #58 the mdbook version was temporary hardcoded to v0.2.1 due to inavailability of the correct files for install.sh. As soon as those are available again this PR needs to be backed out.

PR #205, opened 2019-08-27
Fixup book CI

jamesmunns:
mdbook now has binary releases: https://github.com/rust-lang-nursery/mdBook/releases

Edit:

closes https://github.com/rust-embedded/book/issues/59

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -4,11 +4,10 @@ main() {
     local tag=$(git ls-remote --tags --refs --exit-code \
                     https://github.com/rust-lang-nursery/mdbook \
                         | cut -d/ -f3 \
-                        | grep -E '^v[0.1.0-9.]+$' \
+                        | grep -E '^v[0-9\.]+$' \
                         | sort --version-sort \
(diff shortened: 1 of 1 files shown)


ci/install.sh
```diff
@@ -4,11 +4,10 @@ main() {
     local tag=$(git ls-remote --tags --refs --exit-code \
                     https://github.com/rust-lang-nursery/mdbook \
                         | cut -d/ -f3 \
-                        | grep -E '^v[0.1.0-9.]+$' \
+                        | grep -E '^v[0-9\.]+$' \
```
jamesmunns: Also, this regex didn't work as I assume it was meant to.

If we'd like to constrain to a major version, we should replace this with:

```sh
grep -E '^v0.3.[0-9]+$'
```

jamesmunns: Oh, this implicitly takes us from 0.2.x to 0.3.x. We may want to see if this has negative implications. @steveklabnik this may need to introduce a third version to the upstream book version hack.

PR #212, opened 2019-11-11
Make attributes visible in c-tips chapter

### LINKED ISSUE #211: Examples in intro for C developers lack attributes
It seems like the lines with attributes (i.e. `cfg` switches) are being dropped when rendered:

missing_attr

austinglaser:
Fixes #211

### DIFF
--- a/src/c-tips/index.md
+++ b/src/c-tips/index.md
@@ -44,10 +44,10 @@ Then, in your code, use `#[cfg(feature="FIR")]` to control what is included.
 ```rust
 /// In your top-level lib.rs
 
-#[cfg(feature="FIR")]
+##[cfg(feature="FIR")]
 pub mod fir;
 
-#[cfg(feature="IIR")]
+##[cfg(feature="IIR")]
 pub mod iir;
 ```
 
@@ -77,9 +77,9 @@ for example:
 
 ```rust
 const fn array_size() -> usize {
-    #[cfg(feature="use_more_ram")]
+    ##[cfg(feature="use_more_ram")]
     { 1024 }
-    #[cfg(not(feature="use_more_ram"))]
+    ##[cfg(not(feature="use_more_ram"))]
     { 128 }
 }
 
@@ -328,7 +328,7 @@ fn main() {
 To ensure layouts that are interoperable with C, use `repr(C)`:
 
 ```rust
-#[repr(C)]
+##[repr(C)]
 struct Foo {
     x: u16,
     y: u8,
@@ -348,7 +348,7 @@ fn main() {
 To ensure a packed representation, use `repr(packed)`:
 
 ```rust
-#[repr(packed)]
+##[repr(packed)]
 struct Foo {
     x: u16,
     y: u8,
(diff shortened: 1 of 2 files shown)


andre-richter:
In my browser (Firefox) they are rendered:

image

adamgreig:
I also see them render OK in Chrome; what browser are you using @austinglaser ?

austinglaser:
Firefox 70.0.1, on ubuntu. I don't have any custom styles enabled.

If you build from my branch, does the chapter mis-render on your browsers?

andre-richter:
> Firefox 70.0.1, on ubuntu. I don't have any custom styles enabled.
> 
> If you build from my branch, does the chapter mis-render on your browsers?

Same browser here, to the last digit. Ubuntu 64bit.

austinglaser:
If I look at the page source for one of the offending examples, I see:

```html
<pre><pre class="playpen"><code class="language-rust">
<span class="boring">#![allow(unused_variables)]
fn main() {
</span>/// In your top-level lib.rs

<span class="boring">[cfg(feature=&quot;FIR&quot;)]
</span>pub mod fir;

<span class="boring">[cfg(feature=&quot;IIR&quot;)]
</span>pub mod iir;
<span class="boring">}
</code></pre></pre>
```

What do you see there on your system?

austinglaser:
I see the same behavior (and the same static html, including `boring` spans) when building `master` locally with `mdbook` v0.3.4. Do I need to provide any additional arguments/configuration to `mdbook` other than `mdbook serve` in the root directory?

I've tried explicitly clearing my browser's cache, and I see the same page.

I'm looking at https://rust-embedded.github.io/book/c-tips/index.html. Is there another hosting location for the book that could cause this confusion?

Are there other things I could check? Does `mdbook` use `rustdoc` internally -- could that be a source of version mismatches? The rendering you're seeing seems like it's likely to come from a newer version of the tool which is smarter about attributes -- but the fact that I see this on the public site would seem to indicate that it's unlikely to be solely an issue with tooling versions on my system.

adamgreig:
Ah, I'm looking at https://docs.rust-embedded.org/book/c-tips/index.html which renders OK, whereas your link to rust-embedded.github.io does not render correctly (with the attributes missing).

austinglaser:
There it is! Your link renders fine for me as well.

That probably means we should hold off on merging this until verifying that the changes render correctly when they eventually get pushed to the rust-embedded.org site.

Also, I've found a couple other places in the book where attributes are mistakenly getting swallowed -- I'll add them to this PR as well.

adamgreig:
This repository is configured to serve this repository's `gh-pages` branch at the rust-embedded.github.io link, and the gh-pages clearly has the `boring` spans: https://github.com/rust-embedded/book/blob/gh-pages/c-tips/index.html

Meanwhile the `docs` repository's `gh-pages` branch is hosted at docs.rust-embedded.org and also has CI build this repository, which does _not_ have the `boring` tags: https://github.com/rust-embedded/docs/blob/gh-pages/book/c-tips/index.html

I suspect the problem is that `docs` uses mkbook 0.2.1:

https://github.com/rust-embedded/docs/blob/master/ci/install.sh

while `book` uses latest:

https://github.com/rust-embedded/book/blob/master/ci/install.sh

austinglaser:
I've opened a PR to make the `mdbook`  on `docs` the same as `book`. Additionally, I've fixed what I _think_ are all the occurrences of erroneously-swallowed code in examples (at least `rg boring book` after a build doesn't turn up any offenders

austinglaser:
It looks like this is fixed in mdbook v0.3.5: https://github.com/rust-lang/mdBook/pull/1088

adamgreig:
Thanks for chasing this down, @austinglaser!

PR #219, opened 2020-01-06
Mention discovery book more prominently

phansch:
Hi, I just started reading the book and only by chance found the discovery book via the bookshelf.

I think it's worth mentioning it explicitly somewhere on this page for people who are not familiar with embedded development.


phansch:
I was also thinking of mentioning the discovery book directly in the first paragraph on the page to redirect beginners immediately, but I'm not sure if that's too much?

My understanding of the target audiences of the various books currently is:

1. Discovery book for embedded programming beginners
2. This book for people with embedded experience
3. The embeddonomicon for more advanced things

thejpster:
Thanks for the PR. I agree, the extra pointer to the beginners book will be useful for some.

PR #222, opened 2020-01-27
Replace nursery links with rust-lang links

phansch:
Both repos have been moved to the `rust-lang` org and the links are
still redirecting properly. This way we just avoid the redirect and make
the link shorter.


thejpster:
Thanks for the PR! Looks good to me.

PR #223, opened 2020-01-27
Add a Glossary appendix page

### LINKED ISSUE #220: Including a Glossary
I realize that this book is not targeted to embedded beginners. It would still be nice to have some place to look up common appreviations used in embedded development with a link to more details. Being able to just hit `s` instead of going to search would make things a lot easier. The [rustc-guide](https://rust-lang.github.io/rustc-guide/appendix/glossary.html#appendix-c-glossary) might be a good example.

Some abbreviations from the first few pages that are never explained in the book:

* I2C
* SPI
* USART
* FPU
* PAC

I would be happy to give this a start

phansch:
Same structure as the [rustc-guide] basically.

I added some initial abbreviations and hope the explanations are correct.
I think it would be nice to get this merged without much iteration, so other people can start filling things in, too.

Closes #220

[rustc-guide]: https://rust-lang.github.io/rustc-guide/appendix/glossary.html#appendix-c-glossary

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -46,3 +46,7 @@ more information and coordination
     - [A little Rust with your C](./interoperability/rust-with-c.md)
 - [Unsorted topics](./unsorted/index.md)
   - [Optimizations: The speed size tradeoff](./unsorted/speed-vs-size.md)
+
+---
+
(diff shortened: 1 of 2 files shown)


thejpster:
Good idea, thanks!

PR #231, opened 2020-03-25
Add #![feature(param_attrs)] to enable exception example

bobgates:
On rust V1.39 nightly, at least, the exceptions example gives an error. Adding a line to the example allows it to run.

I'm very green at rust embedded, so there may be better ways, but this makes the example work.


therealprof:
Hi @bobgates, I believe that is a fluke in the nightly build you're using because param_attrs where stabilized around that time, cf. https://github.com/rust-lang/rust/pull/64010.

bobgates:
Okay, with a current Rust nightly 1.44, this problem goes away.

PR #233, opened 2020-04-26
Flip111/patch 8

jamesmunns:
Supersedes #164


jamesmunns:
Hey @adamgreig, this applies your suggestion from https://github.com/rust-embedded/book/pull/164#issuecomment-464517239. I'm not sure why this didn't just update #164, but let's get this merged if we can :)

PR #239, opened 2020-05-17
QEMU debugging updates

### LINKED ISSUE #217: Updated link for Windows OpenOCD download
Hello, I just got started with the book on my Windows machine. The old OpenOCD link leads to something broken and deprecated. I found the following link much more useful in getting OpenOCD installed:

https://gnutoolchains.com/arm-eabi/openocd/

### LINKED ISSUE #234: Need help debugging qemu
I'm not sure if I am missing something or not. I think I don't have everything neccessary installed but cannot figure out what's missing and how to install it.

`src/libcore/slice/mod.rs: No such file or directory.`

Maybe you can point me in the right direction?

```
$ cargo build --target thumbv7m-none-eabi --example hello --release
   Compiling typenum v1.12.0
   Compiling semver-parser v0.7.0
   Compiling proc-macro2 v1.0.10
   Compiling unicode-xid v0.2.0
   Compiling syn v1.0.18
   Compiling cortex-m v0.6.2
   Compiling stable_deref_trait v1.1.1
   Compiling vcell v0.1.2
   Compiling cortex-m-rt v0.6.12
   Compiling cortex-m-semihosting v0.3.5
   Compiling r0 v0.2.2
   Compiling panic-halt v0.2.0
   Compiling semver v0.9.0
   Compiling volatile-register v0.2.0
   Compiling rustc_version v0.2.3
   Compiling quote v1.0.4
   Compiling bare-metal v0.2.5
   Compiling generic-array v0.12.3
   Compiling generic-array v0.13.2
   Compiling as-slice v0.1.3
   Compiling aligned v0.3.2
   Compiling cortex-m-rt-macros v0.1.8
   Compiling app v0.1.0 (/home/fuj/git/home/rust-playground/embedded/app)
    Finished release [optimized + debuginfo] target(s) in 54.48s

$ qemu-system-arm \
                         -cpu cortex-m3 \
                         -machine lm3s6965evb \
                         -nographic \
                         -semihosting-config enable=on,target=native \
                         -gdb tcp::3333 \
                         -S \
                         -kernel target/thumbv7m-none-eabi/debug/examples/hello
Hello, world!
```

When I attach gdb, I get an error that some files are missing:

```
$ arm-none-eabi-gdb -f target/thumbv7m-none-eabi/debug/examples/hello
GNU gdb (GDB) 9.1
<snip>
Reading symbols from target/thumbv7m-none-eabi/debug/examples/hello...
(gdb) target remote :3333
Remote debugging using :3333
0x000012cc in <core::ops::range::Range<usize> as core::slice::SliceIndex<[T]>>::index_mut () at src/libcore/slice/mod.rs:2927
2927    src/libcore/slice/mod.rs: No such file or directory.
(gdb) break main
Breakpoint 1 at 0x444: file examples/hello.rs, line 11.
(gdb) continue
Continuing.

Breakpoint 1, main () at examples/hello.rs:11
/home/fuj/git/home/rust-playground/embedded/app/examples/hello.rs:11:186:beg:0x444
11      #[entry]
(gdb) next
[Inferior 1 (process 1) exited normally]
(gdb) quit
```

### LINKED ISSUE #237: interface/stlink-v2-1.cfg is deprecated, please switch to interface/stlink.cfg
In section 1.4.4, the proposed configuration seems to be deprecated by STM.
`openocd -f interface/stlink-v2-1.cfg -f target/stm32f3x.cfg`
since it generates a warning
`WARNING: interface/stlink-v2-1.cfg is deprecated, please switch to interface/stlink.cfg`
the cfg file proposed by openocd seems to work fine, so perhaps the book could be updated accordingly.
`openocd -f interface/stlink.cfg -f target/stm32f3x.cfg`

BTW, It took me about an hour to get through the introduction/installation chapter. Considering the amount of different SW components involved, that was a smooth ride! Thanks for writing this book!

blelem:
Addressing a few issues:
Fixes #237, Fixes #234, Fixes #217 

The biggest issue this PR is addressing is correcting the "debugging QEMU with GDB" section that really didn't work with latest ARM toolchain (on my setup), because of two things:
 - gdb didn't map the rust libcore functions properly.  For example, after connecting to a QEMU instance, it did not show the current PC to be on the Reset function, but some random method.  
 - gdb didn't step over (`next`), after breaking at `main`. It just ran to the end of the program.

The user is now instructed to disable the LDD and use the GNU ARM linker instead, which addresses the issues with the mapping.

And instead of doing a simple, `break main`  which did not allow the user to step over after hitting the breakpoint, the user is now guided to do a more complicated,
`break hello::__cortex_m_rt_main`. After hitting that breakpoint, the `next` and `step` commands are working as one would expect.

I am no expert in embedded rust, so the proposed solution may not be optimal, but I think they are better than the current instructions provided by the book which are not working at all on my setup. (Windows 10, with latest versions of QEMU, ARM toolchain, Rust)

### DIFF
--- a/src/start/qemu.md
+++ b/src/start/qemu.md
@@ -16,7 +16,10 @@ substitutions.
 ## Creating a non standard Rust program
 
 We'll use the [`cortex-m-quickstart`] project template to generate a new
-project from it.
+project from it. The created project will contain a barebone application: a good
+starting point for a new embedded rust application. In addition, the project will
+contain an `examples` directory, with several separate applications, highlighting
+some of the key embedded rust functionality. 
 
 [`cortex-m-quickstart`]: https://github.com/rust-embedded/cortex-m-quickstart
 
@@ -151,8 +154,14 @@ target = "thumbv7m-none-eabi"    # Cortex-M3
 ```
 
 To cross compile for the Cortex-M3 architecture we have to use
-`thumbv7m-none-eabi`. This compilation target has been set as the default so the
-two commands below do the same:
+`thumbv7m-none-eabi`. That target is not automatically installed when installing
+the Rust toolchain, it would now be a good time to add that target to the toolchain,
+if you haven't done it yet:
+``` console
+$ rustup target add thumbv7m-none-eabi
+```
+ Since the `thumbv7m-none-eabi` compilation target has been set as the default in 
+ your `.cargo/config` file, the two commands below do the same:
 
 ```console
 cargo build --target thumbv7m-none-eabi
@@ -201,8 +210,6 @@ ELF Header:
 
 `cargo-size` can print the size of the linker sections of the binary.
 
-> **NOTE** this output assumes that rust-embedded/cortex-m-rt#111 has been
-> merged
 
 ```console
 cargo size --bin app --release -- -A
@@ -433,6 +440,26 @@ also running the embedded program.
 
 In this section we'll use the `hello` example we already compiled.
 
+Before getting started remotely debugging our app running in QEMU with GDB,
+let's modify our app configuration `.cargo/config`, to enable the GNU ARM linker: 
+
+```toml
+...
+rustflags = [
+  # LLD (shipped with the Rust toolchain) is used as the default linker
+  #"-C", "link-arg=-Tlink.x",  <--- Comment this line
+
+  # if you run into problems with LLD switch to the GNU linker by commenting out
+  # this line
+  "-C", "linker=arm-none-eabi-ld", #<--- Uncomment this line
+```
+We do that, as the default linker (LLD) doesn't always play nice with the GNU ARM GDB.
+Rebuild the application.
+
+```console
+cargo build --example hello
+```
+
 The first debugging step is to launch QEMU in debugging mode:
 
 ```console
@@ -488,11 +515,11 @@ This reset handler will eventually call our main function. Let's skip all the
 way there using a breakpoint and the `continue` command:
 
 ```console
-break main
+break hello::__cortex_m_rt_main
 ```
 
 ```text
-Breakpoint 1 at 0x400: file examples/panic.rs, line 29.
+Breakpoint 1 at 0x410: file examples\hello.rs, line 13.
 ```
 
 ```console
@@ -502,8 +529,8 @@ continue
 ```text
 Continuing.
 
-Breakpoint 1, main () at examples/hello.rs:17
-17          let mut stdout = hio::hstdout().unwrap();
+Breakpoint 1, hello::__cortex_m_rt_main () at examples\hello.rs:13
+13          hprintln!("Hello, world!").unwrap();
 ```
 
 We are now close to the code that prints "Hello, world!". Let's move forward
@@ -512,15 +539,6 @@ using the `next` command.
 ``` console
 next
 ```
-
-```text
-18          writeln!(stdout, "Hello, world!").unwrap();
-```
-
-```console
-next
-```
-
 ```text
 20          debug::exit(debug::EXIT_SUCCESS);
 ```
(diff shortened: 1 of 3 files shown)


src/start/qemu.md
```diff
@@ -433,6 +440,26 @@ also running the embedded program.
...
 In this section we'll use the `hello` example we already compiled.
 
+Before getting started remotely debugging our app running in QEMU with GDB,
+let's modify our app configuration `.cargo/config`, to enable the GNU ARM linker: 
+
+```toml
+...
+rustflags = [
+  # LLD (shipped with the Rust toolchain) is used as the default linker
+  #"-C", "link-arg=-Tlink.x",  <--- Comment this line
```
adamgreig, line 450: It's still required to have `link-arg=-Tlink.x`, otherwise the linker script isn't used at all. The user would have to just add `linker=arm-none-eabi-ld` while keeping `link-arg=-Tlink.x`.

That said, I'm not sure why changing linker helps in this case anyway, which might be worth a little more investigation.

src/start/qemu.md
```diff
@@ -488,11 +515,11 @@ This reset handler will eventually call our main function. Let's skip all the
 way there using a breakpoint and the `continue` command:
 
 ```console
-break main
+break hello::__cortex_m_rt_main
```
adamgreig, line 518: _Technically_, the name `__cortex_m_rt_main` is an implementation detail of cortex-m-rt and probably shouldn't be relied on for this. I'm not sure if there's any other symbol we could use for a breakpoint, though.

adamgreig:
Thanks for this PR! Generally your changes look good but I want to be sure of what's going on before merging the recommendation to use GCC LD or that specific breakpoint name. Others from @rust-embedded/cortex-m might have a better idea...

src/start/qemu.md
jonas-schievink: I've seen a handful of bugs where LLD messes up debuginfo sections, but I don't think we should recommend GNU LD by default because of that. LLD is the default Rust linker for thumb targets.

blelem: I don't know how much the debug info is messed up. The application compiled with LLD works, stepping through the code with gdb works, at least for that minimalist hello world.
So perhaps it's enough for the time being to just signal in the book the bug with smtg like:

> once connected to the remote qemu, gdb should indicate that it is paused at the reset() function, waiting for your input.
> `Reset () at $REGISTRY/cortex-m-rt-0.6.1/src/lib.rs:473`
> 
>  However in some setup, gdb may instead emit some warnings like : `core::num::bignum::Big32x40::mul_small () at src/libcore/num/bignum.rs:254
> 254     src/libcore/num/bignum.rs: No such file or directory.
> `
> You can safely ignore those warnings. 

I can file a bug for that linker issue, I would appreciate your help if you can suggest where the bug should be filed.

src/start/qemu.md
blelem: If exposing that implementation detail is not desired, I believe this slightly more verbose version would work well for first time users:

```console
list main
```
This will show the source code, from the file examples/hello.rs

```text
6       extern crate panic_halt;
7
8       use cortex_m_rt::entry;
9       use cortex_m_semihosting::{debug, hprintln};
10
11      #[entry]
12      fn main() -> ! {
13          hprintln!("Hello, world!").unwrap();
14
15          // exit QEMU
```
We now would like to add a breakpoint just before the "Hello, world!", which is on line 13. We do that with the `break` command:

```console
break 13
```

src/start/qemu.md
blelem: I updated the PR, removing the paragraph recommanding to switch to the GNU LD. 

I added some text hinting at the possible debuginfo mix-up glitch, and that it can be safely ignored.

src/start/qemu.md
blelem: I updated the PR, replacing setting up a breakpoint at the main symbol with a gentle "set your breakpoint on line 13". I think this way of doing will work well for people following this text.

PR #245, opened 2020-06-03
Fix Typos and Improve Readability

lonesometraveler:
First, I want to thank you all for this great resource. I am learning a lot from it. 

I found some typos while reading it for a second time. This PR fixes typos in concurrency/index.md.


src/concurrency/index.md
```diff
@@ -274,7 +274,7 @@ fn timer() {
 We've moved our `unsafe` code to inside our carefully-planned abstraction,
 and now our application code does not contain any `unsafe` blocks.
 
-This design requires the application pass a `CriticalSection` token in:
+This design requires the application to pass a `CriticalSection` token in:
```
adamgreig, line 277:
```suggestion
This design requires that the application passes a `CriticalSection` token in:
```

src/concurrency/index.md
```diff
@@ -509,7 +509,7 @@ Since we can't move the `GPIOA` out of the `&Option`, we need to convert it to
 an `&Option<&GPIOA>` with `as_ref()`, which we can finally `unwrap()` to obtain
 the `&GPIOA` which lets us modify the peripheral.
 
-If we need a mutable references to shared resources, then `borrow_mut` and `deref_mut`
+If we need mutable references to shared resources, then `borrow_mut` and `deref_mut`
```
adamgreig, line 512:
```suggestion
If we need a mutable reference to a shared resource, then `borrow_mut` and `deref_mut`
```

Since we can only have one mutable reference to a given resource, perhaps keeping it in the singular is clearer here.

src/concurrency/index.md
```diff
@@ -383,8 +383,8 @@ to get a safe counter with no unsafe code at all!
 
 This is great for simple types like the `u32` of our counter, but what about
 more complex types which are not Copy? An extremely common example in an
-embedded context is a peripheral struct, which generally are not Copy.
-For that we can turn to `RefCell`.
+embedded context is a peripheral struct, which generally is not Copy.
```
adamgreig, line 386: Thanks, it looks like `is not Copy` is the more commonly used spelling instead of `are not`.

src/concurrency/index.md
lonesometraveler: If we use "that", should not it be "requires that the application pass (not passes)"?

src/concurrency/index.md
lonesometraveler: I just thought it must agree with the singular subject "a peripheral struct".

andre-richter:
Switching assignment with Adam since he's already on top of it.

lonesometraveler:
Hi Andre. 
I thought the original sentence, "This design requires the application pass a `CriticalSection` token in:”, was missing either “that” or “to”. Maybe the author was aware of the subjunctive mood and intentionally omitted “that”. I would suggest “This design requires that the application pass …” or "This design requires the application to pass …”.

src/concurrency/index.md
adamgreig: Sure, I'm happy with "requires that the application pass a..."

lonesometraveler: Thanks. I changed it to "requires that the application pass a...".

Issue #41, opened 2018-10-10
Use Rust 2018 syntax?


thejpster:
Should we be using Rust 2018 syntax in the book? The main change would be removing `extern crate` statements.

hyperslv:
agree but the suggested change has to be synchronized with the appropriate change in source code of the examples at github.com/rust-embedded/cortex-m-quickstart .
I will help & fix both if someone explains me how to synchronize these Repos

therealprof:
What do you mean by synchronize? Open PRs on both and then we'll merge them, couldn't be easier. 😅

hyperslv:
Created a [PR](https://github.com/rust-embedded/book/pull/248) on this issue

hyperslv:
This issue might be closed since the appropriate [PR](https://github.com/rust-embedded/book/pull/248) was merged.

Narukara:
There are still some `extern crate` statements in the book. For example:

https://github.com/rust-embedded/book/blob/19f798d448835a4888e3b3eae7fe69f1d61d8681/src/peripherals/singletons.md?plain=1#L65

https://github.com/rust-embedded/book/blob/09986cd352404eb4659db44613b27cac9aa652fc/src/collections/index.md?plain=1#L33

Shall we fix them all?

eldruin:
@Narukara I would welcome a PR fixing them all. We could then close this issue.

PR #248, opened 2020-06-13
Provide a note on 'extern crate' usage in edition 2018 syntax of Rust

hyperslv:
I was confused reading 'extern crate' in the examples of the book since it looks like old Rust syntax and encounters quite rare nowadays. So it is looked like not-updated. I saw an [issue](https://github.com/rust-embedded/book/issues/41) which tells that other people also had same thoughts. So my change is to make it clear that usage of this declaration form is intended and not-updated.

### DIFF
--- a/src/start/panicking.md
+++ b/src/start/panicking.md
@@ -51,11 +51,11 @@ profile. For example:
 
 // dev profile: easier to debug panics; can put a breakpoint on `rust_begin_unwind`
 #[cfg(debug_assertions)]
-extern crate panic_halt;
+use panic_halt as _;
 
 // release profile: minimize the binary size of the application
 #[cfg(not(debug_assertions))]
-extern crate panic_abort;
+use panic_abort as _;
 
 // ..
 ```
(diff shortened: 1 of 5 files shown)


adamgreig:
I think it would be even better to change the code from `extern crate panic_halt;` to `use panic_halt as _;`

hyperslv:
appropriate changes were made to github.com/rust-embedded/cortex-m-quickstart - [PR](https://github.com/rust-embedded/cortex-m-quickstart/pull/89)

src/start/panicking.md
```diff
@@ -64,6 +64,12 @@ In this example the crate links to the `panic-halt` crate when built with the
 dev profile (`cargo build`), but links to the `panic-abort` crate when built
 with the release profile (`cargo build --release`).
 
+> Note that the _underscore import_ form of the _use_ declaration is used to import a `#[panic_handler]` function
+> symbol without importing of trait's symbol itself since it is not used elsewhere on its own.
+> Sometimes you might see the `extern crate` form of importing which is used for the same reason but it is unidiomatic.
+> `extern crate` should be used only for importing _"sysroot"_ crates (crates distributed with Rust itself) like `proc_macro`,
+> `alloc`, `test`, and similar ones.
+
```
adamgreig, line 72: Thanks for going through and making those changes everywhere! I have a suggestion for the wording of the comment, since I don't think it's right to say "without importing of trait's symbol itself" -- there are no traits involved here.

```suggestion
> The `use panic_abort as _;` form of the `use` statement is used to ensure the `panic_abort` panic handler is
> included in our final executable while making it clear to the compiler that we won't explicitly use anything from
> the crate. Without the `as _` rename, the compiler would warn that we have an unused import.
> Sometimes you might see `extern crate panic_abort` instead, which is an older style used before the
> 2018 edition of Rust, and should now only be used for "sysroot" crates (those distributed with Rust itself) such
> as `proc_macro`, `alloc`, `std`, and `test`.

```

PR #250, opened 2020-06-14
Ease the readers into the Discovery book

### LINKED ISSUE #195: Hardware example inconsistency (STM32F3 versus TM4C123x)
from: https://github.com/rust-embedded/wg/issues/365 (reported by @josephpenafiel)

-------------------------------

I couldn't find any site where I could write the Rust guys directly, so here I am. I just want to point out an inconsistency in this book. If you introduce a specific hardware in this book (the STM32F3 discovery board), why are you writing example code about different hardware??? (tm4c123x, in Section 2 Memory map registers). It's just plain confusing and annoying. 
Stick with st or ti, don't mix things up. I think that just scares people away. I know that this book is aimed at people that have some knowledge about embedded systems in general, but my point is about consistency.
That's just something I wanted to get off my chest. Anyone else agree with me? am I wrong?

Regards!

blelem:
resolves #195

The book has a great and slow introduction to embedded rust using the STM32F3, writing an hello world and then at chapter "Memory-Map registers" takes a hard turn, goes quite hardcore into hardware registers, using other hardware than the STM32F3. That can be quite intimidating for someone new to embedded world, who doesn't quite yet know what systick is, and may just want to blink some leds on his board.

For those Sunday tinkerers that got started with the rust-embedded book that do not yet have the technical chops to go through these details,  I tried to write an escape hatch to the more palatable Discovery Book.
 
I also updated a few bits of code that needed updates to match the latest tools update.


src/start/registers.md
```diff
@@ -15,19 +18,30 @@ You may well find that the code you need to access the peripherals in your micro
...
+## Micro-architecture crate
 
 Let's look at the SysTick peripheral that's common to all Cortex-M based micro-controllers. We can find a pretty low-level API in the [cortex-m] crate, and we can use it like this:
 
 ```rust,ignore
+#![no_std]
+#![no_main]
 use cortex_m::peripheral::{syst, Peripherals};
 use cortex_m_rt::entry;
+extern crate panic_halt;
```
adamgreig, line 40:
```suggestion
use panic_halt as _;
```

adamgreig:
Thanks for this update, looks good!

I wonder if we should link to the newer https://crates.io/crates/stm32f3-discovery crate, which is recently updated, rather than the old `f3` crate, which hasn't been updated in 2 years.

src/start/registers.md
blelem: I've updated all occurrences on the page.

blelem:
Thanks @adamgreig for the quick review!
 
> I wonder if we should link to the newer https://crates.io/crates/stm32f3-discovery crate, which is recently updated, rather than the old `f3` crate, which hasn't been updated in 2 years.

I agree, linking to a maintained crate sounds much better. I've updated my PR accordingly. 
The Discovery book still uses that old `f3` crate, let see how easily I can switch the crate to the stm32f3-discovery crate.

adamgreig:
Ah boo, this is conflicting with https://github.com/rust-embedded/book/pull/248 which made the same `use crate as _` changes. Please could you rebase or merge with master to resolve them?

blelem:
Merge conflict resolved!

PR #251, opened 2020-06-17
Note on transformation of static variables by attribute exception

hyperslv:
Hi All
The "// `COUNT` has type `&mut u32` and it's safe to use" comment confused me since I did not expect that transformation of the source code so deep comparing with the C language.
Finally I found how it (works)[https://docs.rs/cortex-m-rt-macros/0.1.5/src/cortex_m_rt_macros/lib.rs.html#447].
This PR is to provide a note about the transformation.
Thanks, Slava

### DIFF
--- a/src/start/exceptions.md
+++ b/src/start/exceptions.md
@@ -49,7 +49,7 @@ possible.
 
 > Note that the `exception` attribute transforms definitions of static variables
 > inside the function by wrapping them into `unsafe` blocks and providing us
-> with new appropriate variables of type `&mut` which names match user defined.
+> with new appropriate variables of type `&mut` of the same name.
 > Thus we can use operator `*` to access the values of the variables without
 > need to wrap them into the `unsafe` blocks.
 
--- a/src/start/exceptions.md
+++ b/src/start/exceptions.md
@@ -51,7 +51,7 @@ possible.
 > inside the function by wrapping them into `unsafe` blocks and providing us
 > with new appropriate variables of type `&mut` of the same name.
 > Thus we can use operator `*` to access the values of the variables without
-> need to wrap them into the `unsafe` blocks.
+> needing to wrap them in an `unsafe` block.
 
 ## A complete example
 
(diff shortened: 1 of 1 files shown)


hyperslv:
r?

src/start/exceptions.md
```diff
@@ -47,6 +47,12 @@ use `static mut` variables. How is this possible? This is possible because
 `exception` handlers can *not* be called by software thus reentrancy is not
 possible.
 
+> Note that the `exception` attribute makes transformation of static variables
```
therealprof, line 50: This sentence no English. 😅

hyperslv: @therealprof I have changed wording. Please, take a look

src/start/exceptions.md
```diff
@@ -47,6 +47,12 @@ use `static mut` variables. How is this possible? This is possible because
 `exception` handlers can *not* be called by software thus reentrancy is not
 possible.
 
+> Note that the `exception` attribute transforms definitions of static variables
+> inside the function by wrapping them into `unsafe` blocks and providing us
+> with new appropriate variables of type `&mut` which names match user defined.
```
therealprof, line 52:
```suggestion
> with new appropriate variables of type `&mut` of the same name.
```

src/start/exceptions.md
```diff
@@ -47,6 +47,12 @@ use `static mut` variables. How is this possible? This is possible because
...
+> Thus we can use operator `*` to access the values of the variables without
```
therealprof, line 53:
```suggestion
> Thus we can derefence the reference via `*` to access the values of the variables without
```

src/start/exceptions.md
```diff
@@ -47,6 +47,12 @@ use `static mut` variables. How is this possible? This is possible because
...
+> need to wrap them into the `unsafe` blocks.
```
therealprof, line 54:
```suggestion
> needing to wrap them in an `unsafe` block.
```

PR #253, opened 2020-06-20
Note on how to rebuild if memory.x is changed

hyperslv:


therealprof:
I don't think this is true in general. I think we had some previous discussion about this and IIRC the conclusion was: it depends on where the build.rs and memory.x scripts are. I'm okay with the text as such but maybe tone down "does not track" into "may not track"?

hyperslv:
@therealprof, I have updated the Note as you suggested.
Thanks, Slava

PR #261, opened 2020-08-19
Ignore unused argument in closure

GeertJohan:
I'm fairly new to Rust so perhaps I'm overlooking something. But I think this argument should be ignored like the other `_r`'s, right?


eldruin (approved):
Nice catch. Thank you!

bors r+

Issue #263, opened 2020-09-05
New Discovery boards use a stlink-v2-1 and need different openocd config file


Phylex:
Hi in the "Verify Installation" section of the book openocd is started with the interface/stlink.cfg or the /stlink-v2.cfg new boards use a stlink-v2-1 so the correct file is the interface/stlink-v2-1.cfg then it starts without a problem otherwise it errors with
``
Open On-Chip Debugger 0.10.0
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
Info : auto-selecting first available session transport "hla_swd". To override use 'transport select <transport>'.
adapter speed: 1000 kHz
adapter_nsrst_delay: 100
Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
none separate
Info : Unable to match requested speed 1000 kHz, using 950 kHz
Info : Unable to match requested speed 1000 kHz, using 950 kHz
Info : clock speed 950 kHz
Error: open failed
in procedure 'init' 
in procedure 'ocd_bouncer'
``

adamgreig:
Thanks for the report! What version of openocd are you using? In the latest version, there's no `stlink-v2.cfg` or `stlink-v2-1.cfg`, and all STlinks use just `stlink.cfg` instead. However, there hasn't been a release of openocd in a long time, so many users are still on old versions where those other files are required.

Probably the complete solution here is that we'll stop suggesting openocd and suggest probe-rs tools instead, which don't have this sort of issue.

Phylex:
The version is 0.10.0 and is the current packaged version for void-linux.

bitdivine:
Same symptoms, solution and openocd version here.  0.10.0 is also the default version of openocd for Ububtu 20.04.  I note that for programming Blue Pills I also use stlink-v2-1.cfg

julianoes:
This issue seems to have been fixed by #268, however, it would be good to add to the troubleshooting section to try both, if one doesn't work. The F3 I bought seemed to still want `v2`, not `v2-1`.

jeffwhite-619:
I have recently ordered a Discovery board for use in this tutorial and had to change the filename to stlink-v2-1.cfg for it to work. I'm not sure if I received an older board, or if this is the newer board, but it does appear that users like myself may be still be acquiring boards that need this change to proceed through the tutorial.

PR #268, opened 2020-10-09
Fix openocd command in Verify Installation section.

airelil:
For the recent STM32F303VCT6, it should refer interface/stlink-v2-1.cfg.
Same as in https://github.com/rust-embedded/cortex-m-quickstart/blob/be44af69cc91e2ff5b75f400e09488cba44e9f59/openocd.cfg#L7
Otherwise, using the currently specified interface/stlink.cfg produces openocd error:
```
Error: open failed
in procedure 'init'
in procedure 'ocd_bouncer'
```

### DIFF
--- a/src/intro/install/verify.md
+++ b/src/intro/install/verify.md
@@ -17,7 +17,7 @@ ST-LINK header is circled in red.
 Now run the following command:
 
 ``` console
-$ openocd -f interface/stlink.cfg -f target/stm32f3x.cfg
+$ openocd -f interface/stlink-v2-1.cfg -f target/stm32f3x.cfg
 ```
 
 You should get the following output and the program should block the console:


adamgreig:
Thanks for this PR! I think this is related to #263. In the latest version of openocd, you should use the new `stlink.cfg`, which works for all ST-links. However, in openocd 0.10, which is the most recent released version (albeit a few years old), there are still the -v2 etc scripts.

Perhaps the book should be updated to suggest trying out the -v2, especially since lots of people are hitting this issue, but it's not right to only suggest it, since users on the latest versions of openocd (and users in the future) will need what's currently in the book.

airelil:
Fair enough. Thank's for pointing me to #268. It's just from a consistency point of view, openocd version 0.10 is still what user ends up with by following the installation instructions for Linux. Will it be a different version on Mac or Windows platform? Notice also that ver 0.10 is presented in the example of expected output from openocd and st*v2 scripts are referred from openocd.cfg coming with the "cortex-m-quickstart" template recommended by the book in the next section.

Issue #269, opened 2020-10-10
cargo objdump fails on `-triple` flag when used with a recent LLVM 11


airelil:
Running `cargo objdump --bin app --release -- -disassemble -no-show-raw-insn -print-imm-hex` as suggested in [Inspecting](https://rust-embedded.github.io/book/start/qemu.html#inspecting) section of QEMU chapter can produce the following error:
```
cargo objdump --bin app --release -- -disassemble -no-show-raw-insn -print-imm-hex
    Finished release [optimized + debuginfo] target(s) in 0.02s
llvm-objdump: Unknown command line argument '-triple'.  Try: '/home/vk/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/llvm-objdump --help'
llvm-objdump: Did you mean '--triple'?
llvm-objdump: Unknown command line argument '-disassemble'.  Try: '/home/vk/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/llvm-objdump --help'
llvm-objdump: Did you mean '--disassemble'?
llvm-objdump: Unknown command line argument '-no-show-raw-insn'.  Try: '/home/vk/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/llvm-objdump --help'
llvm-objdump: Did you mean '--no-show-raw-insn'?
llvm-objdump: for the --reloc option: may only occur zero or one times!
llvm-objdump: Unknown command line argument '-print-imm-hex'.  Try: '/home/vk/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/bin/llvm-objdump --help'
llvm-objdump: Did you mean '--print-imm-hex'?
```
It seems that it's fixed with https://github.com/rust-embedded/cargo-binutils/commit/aa5464b4eaacfabe808d00db2e155df3740425a2, but might not been yet published on crates.io (https://github.com/rust-embedded/cargo-binutils/issues/93)
I'm using Ubuntu 20.04 with following llvm-objdump version:
```
bin/llvm-objdump --version
LLVM (http://llvm.org/):
  LLVM version 11.0.0-rust-1.47.0-stable
  Optimized build.
  Default target: x86_64-unknown-linux-gnu
  Host CPU: ivybridge
```
[...]

PR #270, opened 2020-10-11
llvm-objdump: Use two hyphens in flags to objdump

adhoore:
LLVM 11 changed the behavior of these tools.
See https://github.com/rust-embedded/book/issues/269

### DIFF
--- a/src/start/exceptions.md
+++ b/src/start/exceptions.md
@@ -249,7 +249,7 @@ If you look at the disassembly of the program:
 
 
 ``` console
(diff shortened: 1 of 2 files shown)


eldruin:
Ouch. This means we would need different instructions depending on the Rust version.

therealprof:
IIRC double hyphen format has been supported for a while in addition to single hyphen. At least that's what I had observed when testing https://github.com/rust-embedded/cargo-binutils/pull/92

eldruin (approved):
Ah, thanks @therealprof and @adhoore.
Looks good to me.

bors r+

Issue #280, opened 2021-02-01
Switch to GitHub Actions for CI


eldruin:
TODO:
- [ ] Switch
- [ ] Fix publication issue about missing `dateutil` python package. See [https://travis-ci.org/github/rust-embedded/book/builds/756864914#L338]

adamgreig:
Done by #285

PR #282, opened 2021-02-08
Fix installing dateutil since it is now a dependency of the GHP import script

eldruin:
This fixes the publication even though still running on Travis CI.
See #280


eldruin:
Pinging @adamgreig since you reviewed the same thing for the discovery book

adamgreig (approved):
bors merge

Thanks!

Issue #283, opened 2021-02-17
Section 1.4.4 references unknown files


aThorp96:
This section tells you to run `$ openocd -f interface/stlink.cfg -f target/stm32f3x.cfg`, except it doesn't specify where to get those files. Are they part of the project template?

adamgreig:
Those files are (or, should be) included with openocd, but if your openocd is a old version (which unfortunately includes the most recent release 0.10.0, from 2017), you might need to try a different filename (`interface/stlink-v2.cfg` perhaps). If you can, I recommend getting the latest nightly version of openocd, there are pre-built images available for various operating systems.

Another option is to use a tool such as PyOCD or [probe-rs](https://probe.rs/), which can also provide a GDB interface, are more up-to-date, and will work with ST-Links as well.

aThorp96:
Thanks for your quick response @adamgreig! I may have those files, but I'm not sure. Do you know where `apt` would install the files?

adamgreig:
You don't need to provide a path, openocd just knows where to look when you pass it those arguments to the `-f` flag. If you have installed openocd from apt you probably don't have stlink.cfg, though, as the apt version is the latest (super old) release.

In Ubuntu they'd be `/usr/share/openocd/scripts/interface/` and `target/`, you'll probably find `stlink-v2.cfg`. (By the way, you can run `locate stlink-v2.cfg` to rapidly find that sort of thing).

aThorp96:
Ah, gotcha. When it didn't find the file I assumed it was a relative path. Thanks! I'll keep the `locate` command in mind :)

aThorp96:
I'm not sure whether or not this warrants any changes to the section, if a you need an open ticket for tracking, so I'll let you close it at your discretion. However my initial issue is resolved. Thanks again for the quick help @adamgreig ! Much appreciated

adamgreig:
Glad to hear it! I think this is already covered by #277 and #263, so I'll close this one. And, uh, maybe make a note in the book this time...

PR #284, opened 2021-02-17
Add note about using more recent openocd interface file. Closes #277 and #263

### LINKED ISSUE #277: 1.4.4. Verify Installation : openOCD - stlink.cfg does not exist [anymore?]
The embedded rust book in section 1.4.4 suggests to use this command to verify the installation:

```
openocd -f interface/stlink.cfg -f target/stm32f3x.cfg
```

However, the `stlink.cfg` file does not exist when installing openOCD (which is version 0.10.0 in my case):

```
$ openocd --version
Open On-Chip Debugger 0.10.0
Licensed under GNU GPL v2
For bug reports, read
        http://openocd.org/doc/doxygen/bugs.html
```

I found the available interface scripts for openOCD at this path (I'm using Manjaro Linux):

```
/usr/share/openocd/scripts/interface
```

The available stlink scripts are:

```
stlink-v1.cfg
stlink-v2.cfg
stlink-v2-1.cfg
```

In my case, using `stlink-v2-1.cfg` solved my problem.

Maybe the Embedded Rust book could elaborate a bit more on using/finding the correct interface file?

adamgreig:
This is a long-standing issue thanks to openocd's confusing rename of this interface file and not having a release for many years (see also #283 and #268).

### DIFF
--- a/src/intro/install/verify.md
+++ b/src/intro/install/verify.md
@@ -20,6 +20,10 @@ Now run the following command:
 $ openocd -f interface/stlink.cfg -f target/stm32f3x.cfg
 ```
 
+> **NOTE**: Old versions of openocd, including the 0.10.0 release from 2017, do
+> not contain the new (and preferable) `interface/stlink.cfg` file; instead you
+> may need to use `interface/stlink-v2.cfg` or `interface/stlink-v2-1.cfg`.
+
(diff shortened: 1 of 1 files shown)


eldruin (approved):
This is great, thank you!

bors r+

PR #288, opened 2021-03-17
Add binary attribute to .jpeg files, closes #287

### LINKED ISSUE #287: *.jpeg should be added as binary files in .gitattributes
Older git versions try to change their line-endings. `.jpg` is already there, but there is currently `src/assets/verify.jpeg` which isn't covered

adamgreig:
Thanks for reporting, @jackh726!


therealprof (approved):
LGTM.

bors r+

PR #292, opened 2021-05-25
Remove $ from cargo-binutils

AlanSwenson:
Currently the copy/paste wont work on this particular one because it copies the `$` as well and then the users console will throw an error


eldruin (approved):
Looks good to me, thanks!

bors r+

PR #294, opened 2021-06-09
Fix punctuation in c-with-rust.md

ziemek99:
I'm not a Rust guy, I was just curious about the topic and I noticed some missing punctuation while reading this book.
This PR adds missing dots at the end of a few sentences. You're doing good work with the book, keep it that way! 😃


eldruin (approved):
Looks good to me. Thanks!

bors r+

PR #308, opened 2021-11-14
added math chapter

### LINKED ISSUE #306: Small sections about performing math functionality
Correct me if I am missing any information, but I was not able to use some standard float operations like `floor` and `ceil` and I required those when calculating register values for a HAL. Am I missing a necessary import for this? I now added the `libm` dependency and was able to use the `floorf` functions for my purposes, but I think not being able to use functions like `floor` is related to having the `#[no_std]`  attribute on the application?

If that is the case, it might be worth an idea to add a small section related to how to perform some basic math operations which are reserved to `std` normally

robamu:
See #306 . I tried to gather some more information about available crates if more complex functionality is required and added links to them as well.

Fixes #306

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -53,6 +53,7 @@ more information and coordination
     - [A little Rust with your C](./interoperability/rust-with-c.md)
 - [Unsorted topics](./unsorted/index.md)
   - [Optimizations: The speed size tradeoff](./unsorted/speed-vs-size.md)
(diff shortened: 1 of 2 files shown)


eldruin (approved):
Great, thank you!
Might be a candidate for inclusion in the microbit rewrite. cc. @hargoniX 

bors r+

hargoniX:
Yeah sooner or later we should have some central resource on some embedded terminology and go to libraries for certain things...

Issue #313, opened 2022-03-08
Segmentation fault


KenxinKun:
Hi! I'm just trying to run the verification steps ([here](https://docs.rust-embedded.org/book/intro/install/verify.html)) and all the variants of the command return me a segmentation fault... Here's the output:

```
$ openocd -f interface/stlink.cfg -f target/stm32f3x.cfg

Open On-Chip Debugger 0.11.0
Licensed under GNU GPL v2
For bug reports, read
	http://openocd.org/doc/doxygen/bugs.html
Info : auto-selecting first available session transport "hla_swd". To override use 'transport select <transport>'.
Info : The selected transport took over low-level target control. The results might differ compared to plain JTAG/SWD
Info : Listening on port 6666 for tcl connections
Info : Listening on port 4444 for telnet connections
Info : clock speed 1000 kHz
[1]    71890 segmentation fault  openocd -f interface/stlink.cfg -f target/stm32f3x.cfg
```

Some extra context:
- STM32F3DISCOVERY connected via USB ST-LINK to miniB to USB-A cable to USB-C adapter to my mac
- Board is functional (i.e. default loaded SW is running with LEDs flashing and gyroscope on etc)
- Running MacOs Monterey 12.2.1
- rustc 1.58.1
- Added rustup target thumbv7em-none-eabihf
- Installed cargo-binutils, llvm-tools-preview and cargo-generate
- gdb 10.3-2021.07
- openocd 0.11.0
- qemu 6.2.0

Any suggestions on how to solve this?

Thanks!

KenxinKun:
Aftes some help from a friend debugging via lldb it seems my issue is the same as the one described [here](https://github.com/libusb/libusb/issues/928).
Uninstalling openocd and reinstalling with `brew install openocd --HEAD` fixed the segmentation fault issue :)
Closing this but adding link above in case someone else gets stuck in the same place!
<img width="833" alt="Screenshot 2022-03-12 at 14 31 41" src="https://user-images.githubusercontent.com/16479107/158022091-1e0434a4-805f-4abb-8c41-dd18d519b598.png">

PR #314, opened 2022-03-16
Update OpenOCD install instruction for #313

neerajadsul:
Hi, I faced same issue as #313. I thought it may help to add the resolution to the docs.


eldruin (approved):
Nice hint, thank you!
bors r+

PR #315, opened 2022-03-17
Correct C interop declarations

### LINKED ISSUE #236: Declaring extern "C" functions to be called from Rust
In [chapter 9.1 "A little C with your Rust"](https://docs.rust-embedded.org/book/interoperability/c-with-rust.html#wrapping-c-functions-and-datatypes), the `extern "C"` function is declared with
```Rust
pub extern "C" fn cool_function(
    i: cty::c_int,
    c: cty::c_char,
    cs: *mut CoolStruct
);
```
I was not able to successfully build my project with such a declaration, I always ended up having "free function without a body" errors. After having compared this approach with [several](https://doc.rust-lang.org/stable/book/ch19-01-unsafe-rust.html?highlight=extern#using-extern-functions-to-call-external-code) [other](https://doc.rust-lang.org/stable/rust-by-example/std_misc/ffi.html) [sources](https://doc.rust-lang.org/nomicon/ffi.html#calling-foreign-functions), it seems to me that the _only_ correct way to declare this function would be
```Rust
extern "C" {
    fn cool_function(
        i: cty::c_int,
        c: cty::c_char,
        cs: *mut CoolStruct
    );
}
```
The `pub extern "C" fn ...` syntax only seems to be valid to expose Rust's functions.

Please correct me if I'm wrong.

Furthermore, the `#[link(name = "<library-name>")]` could be prepended, possibly with a reference to further documentation.

JVMerkle:
Functions with external C linkage have to be wrapped with `extern
"C" { ... }`, `extern "C"` as part of the function signature is not
sufficient.

Resolves #236

Signed-off-by: Julian Merkle <me@jvmerkle.de>

### DIFF
--- a/src/interoperability/c-with-rust.md
+++ b/src/interoperability/c-with-rust.md
@@ -37,11 +37,13 @@ pub struct CoolStruct {
     pub y: cty::c_int,
 }
 
-pub extern "C" fn cool_function(
-    i: cty::c_int,
-    c: cty::c_char,
-    cs: *mut CoolStruct
-);
+extern "C" {
+    pub fn cool_function(
+        i: cty::c_int,
+        c: cty::c_char,
+        cs: *mut CoolStruct
+    );
+}
 ```
 
 Let's take a look at this definition one piece at a time, to explain each of the parts.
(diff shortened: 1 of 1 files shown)


adamgreig (approved):
Thanks! You're right, it's always `extern "C" { pub fn foo(); }` to declare a C function you want to call from Rust like this; it's only `pub extern "C" fn foo(...) { ... }` to define a function in Rust that you wish to be callable from C.

bors r+

PR #317, opened 2022-04-19
Fix use of unaligned references in example

adamgreig:
A recent change in Rust to deny unaligned reference by default (https://github.com/rust-lang/rust/pull/95372) caused the upstream rust build of this book to break: https://github.com/rust-lang/rust/pull/96108#issuecomment-1100646376

This PR removes the unaligned references, instead using `addr_of!()` to get a raw pointer to print.


eldruin (approved):
Thank you for fixing this!
bors r+

PR #321, opened 2022-07-03
"linker `link.exe` not found" fix

tomasApo:
Warned windows users that they need C++ visual studio tools. 
If not installed cargo with complain "linker `link.exe` not found". 
Included /visualstudio.microsoft.com link


eldruin (approved):
Looks good to me, thanks!
bors r+

PR #322, opened 2022-07-09
Fix incorrect explanation

neilpate:
In the description of semihosting is explains that IO occurs on the **host**. This should be **target** surely?

### DIFF
--- a/src/start/semihosting.md
+++ b/src/start/semihosting.md
@@ -1,6 +1,6 @@
 # Semihosting
 
-Semihosting is a mechanism that lets embedded devices do I/O on the host and is
+Semihosting is a mechanism that lets embedded devices do I/O on the target and is
 mainly used to log messages to the host console. Semihosting requires a debug
 session and pretty much nothing else (no extra wires!) so it's super convenient
 to use. The downside is that it's super slow: each write operation can take


hargoniX:
No, semihosting allows the target MCU to perform I/O using the host computer.

neilpate:
@hargoniX do you mean even the peripherals (like DAC etc) are emulated somehow on the host? That is what I interpreted by the statement that semihosting does all the IO on the host. If this is the case then that is really powerful!

hargoniX:
No what the statemt means is that the microcontroller can use the IO capabilities of the host which is basically always just the host's `printf` so you can have debug prints.

neilpate:
Yes that is what I suspected, which is why I suggested the PR in the first place. It is confusing to say "Semihosting is a mechanism that lets embedded devices do I/O on the host and is" as this implies (to me at least) all the IO of the target.

neilpate:
Thanks for the clarification @hargoniX. I understand now and will adjust the text accordingly if you don't mind?

hargoniX:
Sure :+1:

stappersg:
> I understand now and will adjust the text accordingly

_Just do it_

BartMassey:
@neilpate Thanks for your contributions to the discussion. @stappersg Please be kind and polite with new contributors.

I am closing the issue for now. New book development is on <https://github.com/rust-embedded/discovery-mb2>. Can reopen there or here if new wording is proposed.

PR #323, opened 2022-07-17
Updated instructions for running first Hardware example

neilpate:
The code in the example had changed but the instructions and debug output had not. As such, someone experimenting with the example may get confused when there is a mismatch.


eldruin (approved):
Looks good to me, thanks!
bors r+

PR #324, opened 2022-07-19
Improved ligability for hardware.md

tomasApo:
When following Section 2.2 Hardware, I would get stuck trying to launch `gdp`  but wouldn't get stuck in section 2.1 QEMU. There is a disparity between the two sections of code even though they achieving the same function. 
My suggestion is to warn the user about the different versions of `gdp` vs `gdb-multiarch`  in the same fashion as the QEMU section.


eldruin (approved):
Seems sensible to me, thanks!
bors r+

Issue #249, opened 2020-06-14
Fedora has no package named "arm-none-eabi-gdb"


Qwertie-:
The install chapter for linux advises to install this package but it does not exist on fedora 32 or at least not under that name.

Some similar named packages but none look correct

```
================================== Name & Summary Matched: eabi ===================================
arm-none-eabi-binutils-cs.x86_64 : GNU Binutils for cross-compilation for arm-none-eabi target
arm-none-eabi-gcc-cs.x86_64 : GNU GCC for cross-compilation for arm-none-eabi target
arm-none-eabi-gcc-cs-c++.x86_64 : Cross Compiling GNU GCC targeted at arm-none-eabi
arm-none-eabi-newlib.noarch : C library intended for use on arm-none-eabi embedded systems
seabios-bin.noarch : Seabios for x86
======================================= Name Matched: eabi ========================================
seabios.noarch : Open-source legacy BIOS implementation

```

adamgreig:
Perhaps `gdb-multiarch` is available?

Qwertie-:
It is not. Is it possible that it could have been bundled in with the main gdb package? I have been following the book and I don't seem to have had any issues while not having that package.

ryankurte:
not at all familiar with fedora packaging but it looks like it was dropped due to a failed build / missing python dependency?

https://bugzilla.redhat.com/show_bug.cgi?id=1603406

jcrossley3:
Any suggestions for a workaround to follow the book on Fedora 30 or higher? Thanks!

ryankurte:
it appears arm now package arm-none-eabi-gdb in their binary distributions available [here](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm/downloads), if this works for you we could update the book to offer this alternative.

aneonlight:
It seems that this has yet to be fixed in Fedora 35.

I can try and scrape together a pull request. Although, it would break away from the package manager so I'm unsure if a general install vs Fedora specific would be better. (I can do both, just looking for a little guidance)

tylerferrara:
I've been struggling with this and found this process to be the best method forward:

Visit the download page for the latest *arm-gnu-toolchain* -> https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads

```fish
export ARM_TOOLCHAIN="your-arm-gnu-toolchain-url"

# Download and extract to /opt
[...]

PR #335, opened 2022-11-15
doc: Fix `arm-none-eabi-gdb` installation instruction for Fedora 27 or newer to just use `gdb`

mavjs:
- Fixes rust-embedded/book#249
- In the **Rust Embedded Discovery** book, it is noted to use `gdb` instead of `arm-none-eabi-gdb` as pointed out in rust-embedded/discovery#364 that `gdb` on Fedora justworks(tm). This was fixed for _Discovery_ book via rust-embedded/discovery#376
- Likewise, as mentioned per my comment in rust-embedded/book#249, on Fedora 37, `gdb` justworks(tm) as well.

### DIFF
--- a/src/intro/install/linux.md
+++ b/src/intro/install/linux.md
@@ -39,16 +39,13 @@ sudo apt install gdb-arm-none-eabi openocd qemu-system-arm
 
 - Fedora 27 or newer
 
-> **NOTE** `arm-none-eabi-gdb` is the GDB command you'll use to debug your ARM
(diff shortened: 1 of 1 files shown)


eldruin (approved):
Thank you for fixing this!
bors r+

mavjs:
@eldruin Thanks for approving my pull request, however, it seems the checks via bors is pending. Is there anything I should do on my end?

adamgreig:
bors retry

PR #339, opened 2023-01-03
fix: Circled pin header image changed to be better viewed on a b&w e-…

nthState:
# fix: Circled Image not visible on B&W E-Ink screen

## Background

I've been reading this book on my Kindle, which has a black and white e-ink screen, when I came to the section about a pin header, I couldn't easily see the circled pin header.

## Fix

The fix was to change the way the image was highlighted

| Old (Colour) |  Old (Black and White) |
|---|---|
| verify |  verify_bw |

| New (Colour) |  New (Black and White) |
|---|---|]
| verify_new  | verify_bw_new |


adamgreig:
Thanks, this is a good improvement! I wonder if it could be even higher contrast when in colour though -- perhaps the rest of the image could be even further dimmed, or a thick border in another colour added around the highlight region?

jonathanpallant:
I'd merge this as-is; it's a good enough improvement and we can always go around again later.

nthState:
Apologies, I'm not a great user of graphical tools to create what you suggest/need

adamgreig (approved):
Sorry this got dropped, thanks for the improvement!

bors r+

PR #340, opened 2023-01-17
remove extern crate statements

### LINKED ISSUE #41: Use Rust 2018 syntax?

Narukara:
from issue #41 

I kept this one because `alloc` can only be imported that way:

https://github.com/rust-embedded/book/blob/19f798d448835a4888e3b3eae7fe69f1d61d8681/src/collections/index.md?plain=1#L33

Closes #41

### DIFF
--- a/src/collections/index.md
+++ b/src/collections/index.md
@@ -49,9 +49,8 @@ in your program instead of this allocator.
 ``` rust,ignore
 // Bump pointer allocator implementation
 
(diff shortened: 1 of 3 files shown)


eldruin (approved):
Looks good to me, thank you!
bors r+

PR #351, opened 2023-05-06
Update Interoperability section

### LINKED ISSUE #348: Interoperability: On the availability of `std::ffi` and `std::os::raw` in `core`
The [section on interoperability](https://doc.rust-lang.org/stable/embedded-book/interoperability/index.html) mentions that "Neither of these modules [`std::ffi` and `std::os::raw`] are available in `core`". Nonetheless, `core::ffi` has been around since Rust 1.30 and, to my knowledge, contains almost everything that `std::ffi` and `std::os::raw` has to offer.

Moreover, since the documentation of `std::os::raw` explicitly states, "Use `core::ffi` instead" and that `std::ffi` contains everything that `std::os::raw` has, we should probably remove mentions of `std::os::raw`.

abt8601:
Following the discussion on #348, this PR proposes the following changes to the Interoperability section:

- Update descriptions regarding the availability of `core::ffi`.
- Remove mentions of `std::os::raw`.

I intentionally kept mentions of the `cty` crate and the `cstr_core` crate since the next section uses the former. We may also consider updating the following section to use `core::ffi` instead of the `cty` crate.

This PR also introduces two additional changes.

First, it changes the descriptions regarding the implicit conversion between C and Rust types. The book says that the compiler implicitly converts between the two types, but it doesn't sound right to say that the example (reproduced below) works because of the conversion. It works simply because, on platforms where `unsigned int` is 32-bit long, `c_uint` and `u32` are the same type because the former is a type alias of the latter. This PR also removes the `unsafe` from the example since I believe it plays no role.

```rust
unsafe fn foo(num: u32) {
    let c_num: c_uint = num;
    let r_num: u32 = c_num;
}
```

Second, it changes the text formatting in the type correspondence table. The types in the table now use code formatting. Also, the presentation of C pointer types is updated (e.g. `*char` → `char *`).

Closes #348.

### DIFF
--- a/src/interoperability/index.md
+++ b/src/interoperability/index.md
@@ -2,40 +2,42 @@
 
 Interoperability between Rust and C code is always dependent
 on transforming data between the two languages.
-For this purposes there are two dedicated modules
+For this purpose, there is a dedicated module
 in the `stdlib` called
-[`std::ffi`](https://doc.rust-lang.org/std/ffi/index.html) and
-[`std::os::raw`](https://doc.rust-lang.org/std/os/raw/index.html).
+[`std::ffi`](https://doc.rust-lang.org/std/ffi/index.html).
 
-`std::os::raw` deals with low-level primitive types that can
-be converted implicitly by the compiler
-because the memory layout between Rust and C
-is similar enough or the same.
+`std::ffi` provides type definitions for C primitive types,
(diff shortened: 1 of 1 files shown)


eldruin (approved):
Looks good to me, thank you!
bors r+

Issue #350, opened 2023-05-04
Is there a procedure required to translate the repository into another language?


DownyBehind:
I already read #326.

I wonder if it is okay to proceed with the procedure left by @eldruin  on the issue. #326 

I want to translate it into Korean.

thank you.

eldruin:
Yes, that is still the right procedure. We might do something like having `po` files at some point in the future since that approach has now appeared but that requires some time to get right and I for one would not have the time to do it.

DownyBehind:
Thank you for your reply.
Then, as I said, I will add a link after translating it into Korean.

almightychang:
How is it going though? Are you still on translation??

eldruin:
@romancardenas and I are [translating this book into Spanish through transifex](https://explore.transifex.com/rust-embedded/rust-embedded-book/), which works pretty nicely.
Let me know if you want me to start a translation into Korean or any other language.

DownyBehind:
@almightychang I stopped translating for a while because of my other work, but I will do it again soon. But there seems to have been a lot of talk. By the way, are you interested in Korean translation?

@eldruin Do I have to translate in a new way instead of the way you mentioned before (making the translation into github page and adding Link to master branch)?

eldruin:
We are strongly considering taking the approach in #356 and thus we started with a test with a Spanish translation.

DownyBehind:
@eldruin Then can I proceed with the Korean translation in that way? Or should I wait until you guys finish the test?

eldruin:
Sorry @DownyBehind, I wanted to discuss this in our weekly meeting again but have been unable to so far. I will try again tomorrow.
I do not want to waste your time and effort so please hold on for a bit until we try the new approach out.
My status is that we have a roughly 50% Spanish translation [here](https://app.transifex.com/rust-embedded/rust-embedded-book) (about 10% of the snippets are not translatable / not worth translation like console outputs). So far the translation experience has been pretty smooth.

DownyBehind:
@eldruin  Thank you for your answer. Then I'll be waiting for the results of your new attempts

mgeisler:
> My status is that we have a roughly 50% Spanish translation [here](https://app.transifex.com/rust-embedded/rust-embedded-book) (about 10% of the snippets are not translatable / not worth translation like console outputs). So far the translation experience has been pretty smooth.

[...]

Issue #326, opened 2022-08-09
I think this repository can support more languages


Jzow:
have this plan at present?

Jzow:
#126

eldruin:
We would be glad for translations of the book into Chinese and any other language!
[mdbook does not have multilingual support yet](https://github.com/rust-lang/mdBook/issues/5) so I think an easy way to do this for now would be to fork this repository and then translate it. You can then publish it to GitHub pages and we will link it here.
This is what @tomoyuki-nakabayashi has done for the [Japanese version of the discovery book](https://github.com/tomoyuki-nakabayashi/discovery)
Does that sound reasonable to you?
We can help you if some steps do not work.

mgeisler:
Hi there, I wrote mdbook-i18n-helpers and we've been using it for some months now for [Comprehensive Rust 🦀](https://google.github.io/comprehensive-rust/running-the-course/translations.html). I would be happy to help with setting it up for this book.

I see that there are already two hand-written translations available. Getting those converted back to the Gettext PO format will be a non-perfect process. I can try writing a convert tool which takes two Markdown files as input and outputs a PO file. If the translations have kept the Markdown structure intact (same number of paragraphs, same number of list items, ...), it should be possible to convert them back to a PO file.

eldruin:
Sounds great @mgeisler, thank you. It would be great to draw more people into embedded Rust :)
However, please beware that we generally do not have much time to invest, so the setup would need to be easy to manage.

mgeisler:
> Sounds great @mgeisler, thank you. It would be great to draw more people into embedded Rust :)

Yeah, I would hope so!

> However, please beware that we generally do not have much time to invest, so the setup would need to be easy to manage.

Makes sense! In general, the setup I use for the course favors long-term maintainability for some simplicity. The easiest no-setup-needed idea is to clone the repository and do the translations directly inside the English files. However, it's hard to impossible to track changes to the original files that way.

The Gettext approach helps here since it decouples the translation from the source. You extract source text into PO files and translators work on those. When you publish a translation, you read the translation from the PO files. The PO files are small version-controlled databases and there is tooling available which can merge new strings into these databases. This is described in detail in [mdbook-i18n-helpers](https://github.com/google/mdbook-i18n-helpers/).

mgeisler:
> @mgeisler Could you write CI for this repository for `pot` file update and release translated books?

Yes, I would be happy to do this! It might be a week or two before I get to it.

[...]

PR #356, opened 2023-05-16
Use mdbook-i18n-helpers to support translations

mgeisler:
This PR expands the publishing step to also publish a (very partial) Danish translation of the book. The result of this can be seen here: https://mgeisler.github.io/rust-embedded-book/da/intro/index.html.

The infrastructure uses the GNU Gettext tools, as well as a mdbook plugin:

    sudo apt install gettext
    cargo install mdbook-i18n-helpers

The translation lives in po/da.po, which was created with

    MDBOOK_OUTPUT='{"xgettext": {"pot-file": "messages.pot"}}' mdbook build -d po
    msginit -l da -i po/messages.pot -o po/da.po

After creating the da.po file, the translation can be served locally with:

    MDBOOK_BOOK__LANGUAGE=da mdbook serve

The publication flow on GitHub has been expanded to publish translations into subdirectories named after the two-letter ISO_639-1 language codes. The English source text stays the same.

Links to the new translations are added manually in the [index.hbs theme file][1]. I imported the one from mdbook and added a very simple language picker in the top right corner. If someone knows how to extend the existing file without copying it, then I would love to hear about it :smile:

The whole setup tries to be very pragmatic: publishing a language is decoupled from accepting the PO file into the repository, and linking to a translation is decoupled from publishing it.

I replaced the GitHub pages publish step with one that uses a GitHub action instead of a branch — this is not essential, it just matches what I did for my own repository. Feel free to ignore this as you like. Similarly for how the new dependencies are installed and cached: this is just a `cargo install mdbook-i18n-helpers` behind the scenes.

[1]: https://raw.githubusercontent.com/rust-lang/mdBook/master/src/theme/index.hbs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,11 @@ on:
     branches: [ staging, trying, master ]
   pull_request:
 
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
 jobs:
   build:
     runs-on: ubuntu-20.04
@@ -21,25 +26,9 @@ jobs:
       - name: Put pip binary directory into path
         run: echo "~/.local/bin" >> $GITHUB_PATH
 
-      - name: Cache Cargo installed binaries
-        uses: actions/cache@v1
-        id: cache-cargo
+      - uses: moonrepo/setup-rust@v0
         with:
-          path: ~/cargo-bin
-          key: cache-cargo
-      - name: Install mdbook
-        if: steps.cache-cargo.outputs.cache-hit != 'true'
-        uses: actions-rs/install@v0.1
-        with:
-          crate: mdbook
-          version: latest
-      - name: Copy mdbook to cache directory
-        if: steps.cache-cargo.outputs.cache-hit != 'true'
-        run: |
-          mkdir ~/cargo-bin
-          cp ~/.cargo/bin/mdbook ~/cargo-bin
-      - name: Put new cargo binary directory into path
-        run: echo "~/cargo-bin" >> $GITHUB_PATH
+          bins: mdbook, mdbook-i18n-helpers
 
       - name: Build book
         run: mdbook build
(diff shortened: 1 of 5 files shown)


therealprof:
Thanks for the PR. This looks like a great step into the right direction as it is similar to a lot of other projects which have taken the gettext route.

I think we'll need to gauge the interest in the community to help with the translation though since simply having the mechanism to provide all translations from the same repo and maintaining it together is not providing a lot of benefit on it's own and actually increases the maintenance effort required.

Experience from past projects which have taken the exact same route (From English only version over full external translations to gettextized approach) like https://docs.gimp.org/ show that the setup requires a lot of initial work (partially done by this PR), often a restructuring of the existing text to better suit the paraphwise translation mechanism better and a lot of heavy lifting to get a signifcant chunk of translation going before becoming a community driven effort.

Also having the translations in this repo means that now the WG is required to judge the conformance of the offered translations to the community guidelines which might be close to impossible for some languages...

mgeisler:
Hi @therealprof,

> Thanks for the PR. This looks like a great step into the right direction as it is similar to a lot of other projects which have taken the gettext route.

I'm glad you know the system! The infrastructure here is actually the same as what I wrote ~10 years ago for the Mercurial project :smile: As you say, it's a well-known system for software (less so for books).

> I think we'll need to gauge the interest in the community to help with the translation though since simply having the mechanism to provide all translations from the same repo and maintaining it together is not providing a lot of benefit on it's own and actually increases the maintenance effort required.
> 
> Experience from past projects which have taken the exact same route (From English only version over full external translations to gettextized approach) like https://docs.gimp.org/ show that the setup requires a lot of initial work (partially done by this PR), often a restructuring of the existing text to better suit the paraphwise translation mechanism better and a lot of heavy lifting to get a signifcant chunk of translation going before becoming a community driven effort.

I agree, having the infrastructure is only the first step. Actually, people often show up with translations even without having a system in place, as show by all the issues here: https://github.com/rust-lang/book/labels/Translations. For Comprehensive Rust, we had the first translation appear out of the blue about two weeks after we published the course — which prompted me to prioritize the plans for writing a i18n infrastructure.

One step (which I haven't yet done for the Rust course) is to guide people to an online platform for the translations. I'll probably set that up eventually since I hope it will lower the barrier to entry for people.
 
> Also having the translations in this repo means that now the WG is required to judge the conformance of the offered translations to the community guidelines which might be close to impossible for some languages...

Yes, that's very true! For my Rust course, I have the advantage of being able to draw on a vast network of other Googlers, so I can find someone who speaks almost any language :slightly_smiling_face: For other projects, I would apply a simpler standard: let the first person who show up speaking language X be the maintainer of that language. They can then work with other contributors to determine a consistent way to translate the technical terms. Where possible, I've suggested to the translators that they should mimic the [translations of the Rust book](https://doc.rust-lang.org/book/appendix-06-translation.html), but it's not something I'm able to actually check or enforce.

In any case, I hope people can give it a try. If someone (such as @DownyBehind from #350 and @Jzow from #326) wants to play with a Gettext-based translation, then they're free to open PRs against https://github.com/mgeisler/rust-embedded-book. That repository can be a temporary "translation playground" so that you don't have to deal with anything in this repository.

PR #358, opened 2023-06-29
add constgebra to list of math crates

knickish:
`constgebra` is a crate for performing reproducible linear algebra in `const`/`no_std` contexts. Made it for a game, but seems like it might be helpful for embedded also.


eldruin (approved):
Nice, thanks!

PR #361, opened 2023-10-16
Improved hardware.md chapter.

armaxri:
If you follow the documentation blindly, you will end up in a trap with strange compiler warnings. This fix adds a warning about that.

### DIFF
--- a/src/start/hardware.md
+++ b/src/start/hardware.md
@@ -64,6 +64,9 @@ target = "thumbv7em-none-eabihf" # Cortex-M4F and Cortex-M7F (with FPU)
 ```
 
 We'll use `thumbv7em-none-eabihf` as that covers the Cortex-M4F core.
+> **NOTE**: As you may remember from the previous chapter, we have to install
+> all targets and this is a new one. So don't forget to run the installation
+> process `rustup target add thumbv7em-none-eabihf` for this target.
 
 The second step is to enter the memory region information into the `memory.x`
 file.


adamgreig (approved):
Thanks! I think the installation section of the intro chapter does say to use this target for the hardware board, but it's not very clear and a reminder will be helpful.

armaxri:
Actually it uses a different target, which makes everything a little bit confusing. The book talks about the STM32F3 and you have to install the Cortex M3 (the correct) dependencies and in this chapter it jumps to the Cortex M4F...
I'm checking everything right now as an after midnight project, so it will take a few days but maybe there will be a bigger PR later.

adamgreig:
The STM32F3 has a Cortex-M4F CPU core, so `thumbv7em-none-eabihf` is the right target, which is what [this page](https://docs.rust-embedded.org/book/intro/install.html#rust-toolchain) means when it says

> For the STM32F3DISCOVERY board used for the examples in this book, use the thumbv7em-none-eabihf target.

It is a bit confusing that the "STM32F3" has a Cortex-M4F, but it's just how it is - the STM32F4 is also Cortex-M4F, the STM32F1 is Cortex-M3, the STM32H7 is Cortex-M7, but the STM32H5 is Cortex-M33 and so forth.

armaxri:
Thanks a lot! That got me really confused but now I'm back on track! Thanks!

PR #366, opened 2024-01-27
Mention generics in "optimizing dependencies"

Eekle:
I had some friction working with this part of the documentation - I was optimising my dependencies but still getting horrific flash usage.

It turns out to be because the profile override [has subtle interactions with generic code](https://doc.rust-lang.org/cargo/reference/profiles.html#overrides-and-generics).

This PR adds a mention of that footgun.


adamgreig (approved):

adamgreig:
thanks!

PR #368, opened 2024-03-31
Dependencies: changed "qemu-arch-extra" to "qemu-system-arm" on arch section

eramoss:
qemu-arch-extra no longer exists in aur, all packages needed are moved to qemu-system-arm like the other distros


eldruin (approved):
Thank you for fixing this!

PR #369, opened 2024-05-01
Update embedded rust debugging section of tooling.md

CodingInGreen:
Updated tooling.md to include detailed overviews of software like Probe-rs and OpenOCD, and debuggers such as GDB and the Probe-rs extension. Expanded details on hardware probes like Rusty-probe, ST-Link, J-Link, and MCU-Link, highlighting their debugging functionalities and interactions with embedded devices.


eldruin (approved):
Sounds good to me, thank you!

PR #374, opened 2024-08-18
Use aligned address to demonstrate HardFault

decorator-factory:
A volatile read from an unaligned pointer results in this when running in debug mode:
```
panicked at core/src/panicking.rs:221:5:
unsafe precondition(s) violated: ptr::read_volatile requires that the pointer argument is aligned and non-null
```
I changed the address to be aligned (but still invalid on F3 Discovery), so that the expected output is produced.

P.S. I was confused about the lack of output. Maybe default to `panic-semihosting`/`panic-itm` instead of `panic-halt` in these examples?

### DIFF
--- a/src/start/exceptions.md
+++ b/src/start/exceptions.md
@@ -213,7 +213,7 @@ use cortex_m_semihosting::hio;
 fn main() -> ! {
     // read a nonexistent memory location
     unsafe {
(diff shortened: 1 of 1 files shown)


BartMassey (approved):
Thanks for this! Looks good to me.

PR #378, opened 2024-09-19
Add link to Cortex-M comparison to install.md

### LINKED ISSUE #262: Add documentation for Cortex-M23/M33/M35P
_This is sort of a cross-posted issue from https://github.com/rust-embedded/cortex-m-quickstart/issues/94_

I'm getting started with a `ATSAML10xxxx` series project that uses Cortex M23 (ARM v8 baseline)

Since I'm new to Rust development (but already an experienced embedded developer), I've been following [The Embedded Rust Book - Installing the tools](https://docs.rust-embedded.org/book/intro/install.html#installing-the-tools). Specifically it suggests adding the following targets with `rustup target add`

- `thumbv6m-none-eabi`
- `thumbv7m-none-eabi`
- `thumbv7em-none-eabi`
- `thumbv7em-none-eabihf`

As for ARM v8 baseline, mainline and mainline with extension, the targets are respectively:
- `thumbv8m.base-none-eabi` (e.g. Cortex M23)
- ` thumbv8m.main-none-eabi` (Not sure which arch uses this instruction set specifically 🤷‍♀️)
- `thumbv8m.main-none-eabihf`  (e.g. Cortex M33, CortexM35P)

Could these be added to `src/intro/install.md` and also other places I've yet to come across? 

https://github.com/rust-embedded/book/blob/0cd2ca116274b915924c3a7e07c1e046b6f19b77/src/intro/install.md#L14-L40

Maybe a reference to the [Cortex-M comparison table](https://developer.arm.com/ip-products/processors/cortex-m#c-7d3b69ce-5b17-4c9e-8f06-59b605713133) might help with picking out the right instruction set

Feriixu:
Closes #262.

I just stumbled across that issue and thought I'd quickly add this since no one else seemed to want to do it.
I believe this closes said issue, since the rest of the suggested additions in that issue are already there.

### DIFF
--- a/src/intro/install.md
+++ b/src/intro/install.md
@@ -18,6 +18,7 @@ For bandwidth and disk usage concerns the default installation only supports
 native compilation. To add cross compilation support for the ARM Cortex-M
 architectures choose one of the following compilation targets. For the STM32F3DISCOVERY
 board used for the examples in this book, use the `thumbv7em-none-eabihf` target.
+[Find the best Cortex-M for you.](https://developer.arm.com/ip-products/processors/cortex-m#c-7d3b69ce-5b17-4c9e-8f06-59b605713133) 
 
 Cortex-M0, M0+, and M1 (ARMv6-M architecture):
 ``` console


eldruin (approved):
Thank you so much!

PR #382, opened 2024-11-18
Update qemu code slightly to better align with example

diepes:
Compared to code installed by 
cargo generate --git https://github.com/rust-embedded/cortex-m-quickstart

also confirmed at 
https://github.com/rust-embedded/cortex-m-quickstart

Did not sync all the comments just the missing code lines.


adamgreig:
Thanks for the PR! Since you opened this, Rust fixed the bug that made the nop necessary, so this change isn't needed any more; we'll update the quickstart to reflect that.

PR #385, opened 2025-03-11
Provide a better example which shows the usage of singleton!

HosseinAssaran:


HosseinAssaran:
Sure, I’d be happy to! I’ll add some explanatory text and submit a PR to [discovery-mb2](https://github.com/rust-embedded/discovery-mb2) shortly. Thanks for the suggestion!

PR #386, opened 2025-03-11
Clarify usage of #[interrupt] attribute and recommend device crate re…

HosseinAssaran:
…-export

### DIFF
--- a/src/start/interrupts.md
+++ b/src/start/interrupts.md
@@ -19,17 +19,22 @@ The general initialization steps at runtime are always the same:
 * Set the desired priority of the interrupt handler in the interrupt controller
 * Enable the interrupt handler in the interrupt controller
 
-Similarly to exceptions, the `cortex-m-rt` crate provides an [`interrupt`]
-attribute to declare interrupt handlers. The available interrupts (and
-their position in the interrupt handler table) are usually automatically
-generated via `svd2rust` from a SVD description.
+Similarly to exceptions, the cortex-m-rt crate exposes an [`interrupt`] attribute for declaring interrupt handlers. However, this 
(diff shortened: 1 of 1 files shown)


BartMassey:
Thanks much for this! The text is a definite improvement. I wonder why the code needed to change?

Any chance you could do a similar PR for https://github.com/rust-embedded/discovery-mb2

HosseinAssaran:
It demonstrates how to use interrupts and highlights the need to include use device::interrupt. Additionally, since the emulator in this book is based on the [LM3S6965](http://www.ti.com/product/LM3S6965), which does not have a TIM2 peripheral but instead provides TIMER2A and TIMER2B, the code must be adapted accordingly.

src/start/interrupts.md
```diff
@@ -19,17 +19,22 @@ The general initialization steps at runtime are always the same:
 * Set the desired priority of the interrupt handler in the interrupt controller
 * Enable the interrupt handler in the interrupt controller
 
-Similarly to exceptions, the `cortex-m-rt` crate provides an [`interrupt`]
-attribute to declare interrupt handlers. The available interrupts (and
-their position in the interrupt handler table) are usually automatically
-generated via `svd2rust` from a SVD description.
+Similarly to exceptions, the cortex-m-rt crate exposes an [`interrupt`] attribute for declaring interrupt handlers. However, this 
+attribute is only available when the device feature is enabled. That said, this attribute is not intended to be used directly—doing
```
therealprof, line 23:
```suggestion
attribute is only available when the device feature is enabled. That said, this attribute is not intended to be used directly — doing 
```

adamgreig (approved):

PR #390, opened 2025-04-20
Explain why we should avoid static mutable variable and introduce interior mutability

### LINKED ISSUE #389: Discourage use of static mut in singleton
The code in this part about Singleton no longer works as of Rust 2024

https://docs.rust-embedded.org/book/peripherals/singletons.html

```rust
static mut PERIPHERALS: Peripherals = Peripherals {
    serial: Some(SerialPort),
};
```

giving error instead of warning "mutable reference to mutable static"

EvansJahja:
Closes #389


src/peripherals/singletons.md
```diff
@@ -23,6 +23,8 @@ fn main() {
 
 But this has a few problems. It is a mutable global variable, and in Rust, these are always unsafe to interact with. These variables are also visible across your whole program, which means the borrow checker is unable to help you track references and ownership of these variables.
 
+Up until Rust 2024, the usage of static **mutable** variable has been discouraged. Since Rust 2024, static mutable variable results in an error. Of course, there are legitimate use case of having a shared mutable state, it's just that Rust want you to really think about the scope of the mutability. We'll talk about this later when we talk about interior mutability.
```
jannic, line 26: When read literally, "Since Rust 2024, static mutable variable results in an error." is wrong.

A static mutable variable in itself is not an issue at all, if you are comfortable with unsafe code: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a9bb1d76aa1ef7cb07442d890438cd7d

What's discouraged is using references to static mutable variables:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7b96c2c8023526d86d8f9b619eda9846
And even there, it's not a hard error, but a `deny` lint, that can be overridden (`#[allow(static_mut_refs)]`)

Perhaps make that "Since Rust 2024, references to static mutable variables are denied by default."?

src/peripherals/using-interior-mutability.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+where Sync does absolutely nothing.
+
+Another way is to use Rust's `SyncUnsafeCell` which is currently only available in nightly under the flag `#![feature(sync_unsafe_cell)]`.
+
+
+Of course, none of the above are safe. In order to get a proper interior mutability that is safe, we should implement `Sync` ourself and put in our synchronization primitive specific to the hardware such as locking the resource with atomic swap and guarding the code with interrupt-free section (disable interrupt, run some code, enable interrupt), also known as critical section.
+
+You could check how [rust-console gba](https://github.com/rust-console/gba/blob/6a3fcc1ee6493a499af507f8394ee542500721b7/src/gba_cell.rs#L63) implement it for reference.
```
jannic, line 94: Isn't this example a bit too specific and complicated?
I don't have a better example at hand though. Not a big deal, if someone has a better suggestion, it can easily be replaced later.

src/peripherals/singletons.md
EvansJahja: Updated in d918f02c37d2a487610f3f1aecb22e2f20209b9a, thank you for the correction

src/peripherals/using-interior-mutability.md
EvansJahja: agree, let's keep this for now

src/peripherals/using-interior-mutability.md
```diff
@@ -0,0 +1,96 @@
...
+
+* the variable `foo` is moved to another place. Subsequent reference of `foo` would be a compile failure as the value has been moved. 
+* some number of `&foo` borrows, which allow read-only access.
+* a single (exclusive) `&mut foo` mutable borrow, which allow read/write. 
+
+One consequence of a `static` variable is that it cannot be moved, as it lives _statically_ in memory, meaning it is in a fixed place.
+
+A `static` variable can be borrowed as `&foo` without much issue as it is just reading.
+
+The problem occurs when we try to borrow is as mutable. Consider the following:
```
eldruin:
```suggestion
The problem occurs when we try to borrow it as mutable. Consider the following:
```

src/peripherals/using-interior-mutability.md
```diff
@@ -0,0 +1,96 @@
...
+
+Also, the use of `static mut` variables _might_ be deprecated in the future. There's a strong indication of this and having the usage `deny` by default since Rust 2024 is one of them.
+
+## So what is interior mutability?
+
+Interior mutability is a way to "trick" the borrow checker to mutate a borrowed reference `&foo` (notice the lack of `&mut`). This is basically us saying to the compiler "I'm going to mutate the shared reference anyway, please don't stop me.".
+
+Of course, mutating something that is borrowed in other places are undefined behaviors (UB), yet there is a way of doing so, using an `UnsafeCell`. As the name implies, you need to use the `unsafe` keyword when using it.
```
eldruin:
```suggestion
Of course, mutating something that is borrowed in other places is undefined behavior (UB), yet there is a way of doing so, using an `UnsafeCell`. As the name implies, you need to use the `unsafe` keyword when using it. 
```

src/peripherals/using-interior-mutability.md
```diff
@@ -0,0 +1,96 @@
...
+
+You may think that because you're in an embedded programming land and you only have single core, you don't have to think about race condition, but even with a single core CPU without making any thread, you may have interrupts and those interrupt requests may access the shared mutable variable `flag`. The issue here is that the compiler cannot prohibit you from taking multiple mutable reference to the shared mutable variable at compile time, especially since it does not know about the order of executions of threads or interrupts, for example.
+
+Also, the use of `static mut` variables _might_ be deprecated in the future. There's a strong indication of this and having the usage `deny` by default since Rust 2024 is one of them.
+
+## So what is interior mutability?
+
+Interior mutability is a way to "trick" the borrow checker to mutate a borrowed reference `&foo` (notice the lack of `&mut`). This is basically us saying to the compiler "I'm going to mutate the shared reference anyway, please don't stop me.".
```
eldruin:
```suggestion
Interior mutability is a way around the borrow checker to mutate a borrowed reference `&foo` (notice the lack of `&mut`). This is basically us saying to the compiler "I'm going to mutate the shared reference anyway, please don't stop me.".
```
I think "trick" sounds a bit too alluring :) Since we do not want people to use unsafe more that really strictly necessary, I would make the wording slightly more neutral.

src/peripherals/using-interior-mutability.md
```diff
@@ -0,0 +1,96 @@
...
+where Sync does absolutely nothing.
+
+Another way is to use Rust's `SyncUnsafeCell` which is currently only available in nightly under the flag `#![feature(sync_unsafe_cell)]`.
+
+
+Of course, none of the above are safe. In order to get a proper interior mutability that is safe, we should implement `Sync` ourself and put in our synchronization primitive specific to the hardware such as locking the resource with atomic swap and guarding the code with interrupt-free section (disable interrupt, run some code, enable interrupt), also known as critical section.
+
+You could check how [rust-console gba](https://github.com/rust-console/gba/blob/6a3fcc1ee6493a499af507f8394ee542500721b7/src/gba_cell.rs#L63) implement it for reference.
+
+Interior mutability and static mutability is a rather deep topic.  I strongly recommend reading more about static mutable references from [the rust edition guide here](https://github.com/rust-lang/edition-guide/blob/master/src/rust-2024/static-mut-references.md).
```
eldruin:
```suggestion
Interior mutability and static mutability is a rather deep topic.  I strongly recommend reading more about static mutable references from [The Rust Edition Guide here](https://doc.rust-lang.org/edition-guide/rust-2024/static-mut-references.html).
```

eldruin (commented):
Great work! Thank you so much!
I just did some wording adjustments and added some links.
Once that is settled, I would merge it.

PR #391, opened 2025-05-20
portability: add reference to embedded-hal docs

orangecms:
Same as in src/start/registers.md, for every occurrence.


therealprof (approved):
LGTM

PR #395, opened 2025-06-17
Fix incorrect type for semihosted stdout stream #394

### LINKED ISSUE #394: start/exceptions "A complete example" uses incorrect type for semihosted stdout stream
In [Exceptions] sub-chapter of Getting Started, the complete example uses incorrect type for the semihosted stdout stream.

The example uses use hio::HStdout instead of [hio::HostStream] which should be used when using cortex-m-semihosting = "0.5" per manifest included in the cortex-m-quickstart template.

[Exceptions]: https://github.com/rust-embedded/book/blob/10fa1e084365f23f24ad0000df541923385b73b6/src/start/exceptions.md
[hio::HostStream]: https://docs.rs/cortex-m-semihosting/0.5.0/cortex_m_semihosting/hio/struct.HostStream.html

matej-almasi:
This fixes #394.

### DIFF
--- a/src/start/exceptions.md
+++ b/src/start/exceptions.md
@@ -81,7 +81,7 @@ use cortex_m::peripheral::syst::SystClkSource;
 use cortex_m_rt::{entry, exception};
 use cortex_m_semihosting::{
     debug,
-    hio::{self, HStdout},
+    hio::{self, HostStream},
 };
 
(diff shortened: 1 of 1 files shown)


BartMassey (approved):
Haven't tried compiling, but looks good to me!

PR #399, opened 2025-09-07
Add Russian translation repository

THELoer:


BartMassey:
Wow! That looks amazing. Thank you very much.

I think we'll need to find an independent Russian/English speaking reviewer for this.

adamgreig:
Thank you! For translations we ask people to add a link to our copy of the book here:

https://docs.rust-embedded.org/book/index.html#translations

https://github.com/rust-embedded/book/blob/master/src/intro/index.md#translations

We're happy to link to it, but without the ability to review the translation ourselves (and to allow us to keep updating the English copy as required) we don't host the translated one in our own repository.

PR #403, opened 2026-01-19
Update outdated qemu documentation

crespum:
I've seen #400 PR but I went ahead and modified the docs to use the new suggested template (https://github.com/knurling-rs/app-template).

The tricky part is that the new template uses `defmt`, so the `qemu-system-arm` binary is not enough but Ferrous System's [`qemu-run`](https://github.com/knurling-rs/defmt/tree/main/qemu-run/) tool is required to decode the output properly.

I'm not a big fan of third-party dependencies in the official docs, but feedback is welcome.

### DIFF
--- a/src/start/qemu.md
+++ b/src/start/qemu.md
@@ -359,6 +359,9 @@ and RTT to print text.
 
 [defmt]: https://defmt.ferrous-systems.com/
 
+> **NOTE** `defmt` is a third-party dependency (i.e. non-core) widely used in the
+> Embedded Rust ecosystem.
+
 In order to read and decode the messages produced by `defmt` in the host, we need to
 switch the RTT transport output to semihosting. When using real hardware this requires
 a debug session but when using QEMU this Just Works.


src/start/qemu.md
```diff
@@ -311,55 +354,53 @@ HardFault:
 ## Running
 
 Next, let's see how to run an embedded program on QEMU! This time we'll use the
-`hello` example which actually does something.
+`hello` example which actually does something. By default, this example uses [defmt]
```
eldruin:
```suggestion
`hello` example which actually does something. By default, this example uses `[defmt]`
```

src/start/qemu.md
```diff
@@ -311,55 +354,53 @@ HardFault:
...
+and RTT to print text.
 
-For convenience here's the source code of `examples/hello.rs`:
+[defmt]: https://defmt.ferrous-systems.com/
 
-```rust,ignore
-//! Prints "Hello, world!" on the host console using semihosting
+In order to read and decode the messages produced by defmt in the host, we need to
```
eldruin:
```suggestion
In order to read and decode the messages produced by `defmt` in the host, we need to
```

eldruin (commented):
This sounds good to me, thank you!
The dependency is unfortunate but I see the convenience in using the template for a hello world.
Any objections?

crespum:
I've added a note about `defmt` being a third-party dependency. Let me know if I should keep it or drop the commit.

PR #409, opened 2026-03-11
Update teaching material duration for bare-metal Rust

mas192:
Updated the duration of the teaching material for bare-metal Rust development from 1-day to 4-day.


BartMassey (approved):
Thank you!

PR #411, opened 2026-04-10
remove IO chapter, is covered in next chapter

robamu:


BartMassey (commented):
I think this section stub might be intended to be about external I/O rather than generic peripheral stuff. Hard to tell given the stub. It seems reasonable to either remove the stub section or fill it in with something, but not sure which is better.

robamu:
Hmm that is what I thought as well but I think the knowledge from generic peripheral stuff can easily be transferred to external I/O, which usually also invovles writing to and reading from registers.

BartMassey:
I agree that it is hard to tell what is intended. Maybe a discussion of the embedded `Read` and `Write` traits or something?

9names:
The stub has been present for 8 years without being filled in.
Lack of context around what should replace it is a large part of that.

The original author who put the placeholder there does not actively participate in this community any more, and while they are contactable I think this isn't a question worth asking them - we should be filling the book with things we think are important *now*, not trying to peer into the past to try to understand a placeholder's intent.

BartMassey (approved):
Let's do this.

PR #413, opened 2026-06-09
Fixing typo error

m00sp:


eldruin (commented):
Thank you for your suggestion but the command is correct as it is. We want to execute as little possible here. There is [more info here](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Continuing-and-Stepping.html)

